PR #2, opened 2023-01-16
Add readme for rust core

sighphyre:
Adds a basic README to the Rust core, this is necessary because when we publish, the README here will be published to crates.io


chriswk (approved):
Nice start :)

PR #3, opened 2023-01-16
add in a fetch for the client spec in the gh workflow

sighphyre:
There might be a much better way of doing this. I'm open to suggestions. But ultimately, we need to resolve the client spec before we run tests

### DIFF
--- a/.github/workflows/sarif-and-test.yaml
+++ b/.github/workflows/sarif-and-test.yaml
@@ -51,8 +51,12 @@ jobs:
           category: clippy
           wait-for-processing: true
       - name: Get client spec
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/sarif-and-test.yaml
```diff
@@ -50,6 +50,12 @@ jobs:
           sarif_file: rust-clippy-results.sarif
           category: clippy
           wait-for-processing: true
+      - name: Get client spec
+        uses: actions/checkout@v3
+        with:
+          repository: Unleash/client-specification
+          branch: v4.2.2
```
chriswk, line 57:
```suggestion
          ref: v4.2.2
```

chriswk (changes requested):
You need to change to ref, rather than branch

chriswk (approved):

PR #5, opened 2023-01-17
add robustness handling to context property parsing

sighphyre:
Adds some robustness to the context property parsing code. This will now handle cases where the properties object coming from JSON serialisation or FFI layers that use the same have a defined but null field in the context object


chriswk (approved):
Cool. TIL about remove_null_properties

PR #6, opened 2023-01-17
Remove reqwest dependency

sighphyre:
Removes reqwest, we don't actually use it, this was an experiment that got removed without removing the dep


chriswk (approved):
LGTM

PR #7, opened 2023-01-18
feat: improve sad path handling of variants

sighphyre:
This PR solves two things: first it allows us to pass through the name of the toggle to the rule engine in cases where group id is not specified (this is massively an edge case since this is always defined in Unleash). The second is that variant stickiness is now no longer allowed to unwrap a potentially empty property, this now either resolves stickiness correctly or falls back to a random number


unleash-yggdrasil/src/state.rs
```diff
@@ -15,6 +15,32 @@ pub struct InnerContext {
     pub properties: Option<HashMap<String, String>>,
 }
 
+pub struct EnrichedContext {
```
sighphyre, line 18: I have mixed feelings about having a secondary context struct with enriched properties. It looks a bit ugly but it buys us something very, very nice - the language specific SDK code is now shielded from changes in the enriched context which means internal implementations and extensions can happen without impacting the langauge SDK code

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -302,14 +302,15 @@ fn rollout_constraint(mut node: Pairs<Rule>) -> RuleFragment {
                 }
                 custom_stickiness
             }
-            None => context.user_id.clone().or_else(||context.session_id.clone()),
+            None => context
+                .user_id
+                .clone()
+                .or_else(|| context.session_id.clone()),
```
chriswk, line 308: Isn't this `context.user_id.or(context.session_id)`

sighphyre: It's an Option<&T> at this point, so it needs to be `context.user_id.clone().or(context.session_id.clone())`.

That is better though, nice suggestion, it'll come up with the test

sighphyre: Wait, no, clippy doesn't like it, fair comment from clippy - the `or` may not be evaluated when you think it is

unleash-yggdrasil/src/lib.rs
```diff
@@ -122,15 +128,27 @@
...
+                    let target = match get_custom_stickiness(variants, context) {
+                        Ok(stickiness) => stickiness
+                            .or_else(|| {
+                                context
+                                    .user_id
+                                    .clone()
+                                    .or_else(|| context.session_id.clone())
+                            })
+                            .map(|stickiness| {
+                                normalized_hash(&toggle.name, &stickiness, total_weight as u32)
```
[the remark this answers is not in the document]
sighphyre: Fixed

chriswk (approved):
LGTM

PR #8, opened 2023-01-18
feat: Add unicode support for the rule parser

sighphyre:
Extends the rule grammar to tolerate unicode characters and a limited set of punctuation in the string atoms


nunogois (approved):
🚀😎👍 - Teenager characters meaning this LGTM!

PR #10, opened 2023-01-20
ci: update workflow to use rust-cache

chriswk:
We've struggled with buildtimes for our rust-actions. After finding the rust-cache action it seems like we've managed to cache enough of the target folder that we no longer have to build all our dependencies each time we make changes to our own code.

This PR switches this project's workflow to use rust-cache as well


sighphyre (approved):
Pure magic, love it

PR #15, opened 2023-01-25
chore: add basic benchmarks

sighphyre:
Adds basic benchmarks for a sniff test for performance. This isn't a complete analysis of the performance profile, only a rough guide


unleash-yggdrasil/src/lib.rs
```diff
@@ -79,17 +79,12 @@ impl<'de> de::Deserialize<'de> for IPAddress {
     }
 }
 
+#[derive(Default)]
```
sighphyre, line 82: Not relevant to this PR but Clippy is cranky

chriswk (approved):
Cool. Love me some benchmarks.

PR #16, opened 2023-01-25
feat: allow the engine to report the version of the Unleash client spec it support

sighphyre:


chriswk (approved):
Lovely. This will smarten our edge client :)

PR #17, opened 2023-01-25
fix: upgrade user id strategy to string types rather than numerics

sighphyre:
User ids can be string types, prior to this, Yggdrasil was trying to upgrade these to numeric rules like so:

`user_in IN [1,2,3]`

This doesn't work when the user supplies a user id which doesn't parse as a numeric. This PR changes this so that the strategy is upgraded to a string type like so:

`user_in IN ["1","2","3"]`

### DIFF
--- a/unleash-yggdrasil/src/strategy_parsing.rs
+++ b/unleash-yggdrasil/src/strategy_parsing.rs
@@ -689,6 +689,8 @@ mod tests {
     #[test_case("user_id in [1, 3, 6]", true)]
     #[test_case("user_id in [1, 3, 5]", false)]
     #[test_case("user_id not_in [1, 3, 5]", true)]
+    #[test_case("user_id in [\"dfsfsd\"]", false)]
+    #[test_case("user_id not_in [\"dfsfsd\"]", true)]
     fn run_numeric_list_test(rule: &str, expected: bool) {
         let rule = compile_rule(rule).expect("");
         let context = context_from_user_id("6");
--- a/unleash-yggdrasil/src/strategy_upgrade.rs
+++ b/unleash-yggdrasil/src/strategy_upgrade.rs
@@ -101,7 +101,7 @@ fn upgrade_user_id_strategy(strategy: &Strategy) -> String {
     match strategy.get_param("userIds") {
         Some(user_ids) => {
             let user_ids = user_ids
-                .split(",")
+                .split(',')
                 .map(|id| format!("\"{}\"", id.trim()))
                 .collect::<Vec<String>>()
                 .join(",");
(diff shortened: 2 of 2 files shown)


unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -689,6 +689,8 @@ mod tests {
     #[test_case("user_id in [1, 3, 6]", true)]
     #[test_case("user_id in [1, 3, 5]", false)]
     #[test_case("user_id not_in [1, 3, 5]", true)]
+    #[test_case("user_id in [\"dfsfsd\"]", false)]
```
sighphyre, line 692: The grammar supports this case out the box but it's good to layer a test over this so we know for future

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -689,6 +689,8 @@ mod tests {
...
+    #[test_case("user_id not_in [\"dfsfsd\"]", true)]
```
gastonfournier, line 693: Just a minor detail: with this is not clear why `user_id not_in [\"dfsfsd\"]` should be true without looking at the body of the test. I don't know if it's worth adding the `6` parameter from where you're taking the context. 

Something like:
1. scenario: given the context for user id 6
2. when: user_id in [\"dfsfsd\"]
3. then: should be false

gastonfournier (approved):
Looks fine! Just a small detail that you can ignore (all other tests in this file works similarly)

Also, I noticed that Rust files tend to be really large due to embedded tests

unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -99,7 +99,14 @@
 
 fn upgrade_user_id_strategy(strategy: &Strategy) -> String {
     match strategy.get_param("userIds") {
-        Some(user_ids) => format!("user_id in [{}]", user_ids),
+        Some(user_ids) => {
+            let user_ids = user_ids
+                .split(",")
```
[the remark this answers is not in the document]
sighphyre: Fixed

unleash-yggdrasil/src/strategy_parsing.rs
sighphyre: > Just a minor detail: with this is not clear why user_id not_in [\"dfsfsd\"] should be true without looking at the body of the test. I don't know if it's worth adding the 6 parameter from where you're taking the context.

Ah. Yeah fair enough. I've gotten too used to reading these. Probably not a thing I want to do in this PR but a big refactor of the test names might be worthwhile now that I'm no longer the only person poking

> Also, I noticed that Rust files tend to be really large due to embedded tests

Yeah, you get used to it over time. You start to look for the mod test and mark that in your mind as the relevant test code. Takes a bit of time but I prefer it now

gastonfournier: > Yeah, you get used to it over time. You start to look for the mod test and mark that in your mind as the relevant test code. Takes a bit of time but I prefer it now

Challenge accepted :rofl:

PR #19, opened 2023-01-26
fix: handle empty variants

sighphyre:
Handle cases where get_variant is called and variants is Some but empty


chriswk (approved):
Makes sense :)

PR #25, opened 2023-02-13
feat: implement metrics gathering for core

sighphyre:
This adds metrics gathering for the core Yggdrasil client, using atomic primitives behind the scenes, which means this is should still be Send & Sync (yay!). The tradeoff here is that the client won't hold metrics longer than the engine holds the internal state - when passing in a new state for compilation, the engine will return unhandled metrics, that's now the callers responsibility to handle unsent metrics

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::net::IpAddr;
+use std::sync::atomic::AtomicU32;
 #[macro_use]
 extern crate lazy_static;
 #[macro_use]
@@ -11,14 +12,17 @@ pub mod state;
 pub mod strategy_parsing;
 pub mod strategy_upgrade;
 
+use chrono::{DateTime, Utc};
 use error::YggdrasilError;
 use rand::Rng;
 use serde::{de, Deserialize, Serialize};
 use state::EnrichedContext;
+use std::sync::atomic::Ordering;
 use strategy_parsing::{compile_rule, normalized_hash, RuleFragment};
 use strategy_upgrade::upgrade;
 pub use unleash_types::client_features::Context;
-use unleash_types::client_features::{ClientFeatures, Payload, Segment, Variant};
+use unleash_types::client_features::{ClientFeatures, Override, Payload, Segment, Variant};
+use unleash_types::client_metrics::{MetricBucket, ToggleStats};
 
 pub type CompiledState = HashMap<String, CompiledToggle>;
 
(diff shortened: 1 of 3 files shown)


unleash-yggdrasil/src/lib.rs
```diff
@@ -28,7 +32,47 @@ pub struct CompiledToggle {
...
+    pub variants: Vec<CompiledVariant>,
+    pub yes: AtomicU32,
+    pub no: AtomicU32,
+    pub default_variant: AtomicU32,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CompiledVariant {
+    pub name: String,
+    pub weight: i32,
```
nunogois, line 44: I'm curious about this type. Can weight be negative and is it only int?

unleash-yggdrasil/src/lib.rs
```diff
@@ -94,84 +158,144 @@ impl EngineState {
...
+            })
+            .unwrap_or_default()
+    }
+
+    fn resolve_variant<'a>(
+        &self,
+        toggle: &CompiledToggle,
+        variants: &'a Vec<CompiledVariant>,
+        context: &Context,
+    ) -> Option<&'a CompiledVariant> {
```
nunogois, line 231: From a quick search does `'a` here mean that our result type only lives as long as what we're taking in as our parameters? Would love a ELI5 in this context if possible, thanks!

unleash-yggdrasil/src/lib.rs
```diff
@@ -94,84 +158,144 @@ impl EngineState {
...
+        if variants.is_empty() {
+            return None;
+        }
+        if let Some(found_override) = check_for_variant_override(variants, context) {
+            return Some(found_override);
+        }
+        let total_weight: u32 = variants.iter().map(|var| var.weight as u32).sum();
```
nunogois, line 238: Seems like we're `as u32`ing the weights here. Related to https://github.com/Unleash/yggdrasil/pull/25/files#r1104638297?

nunogois (approved):
LGTM

unleash-yggdrasil/src/lib.rs
sighphyre: So this is Yggdrasil being sandwiched between two competing types. Unleash-types passes weight as an i32 and our hashing library requires a u32. To use weight there needs to be a cast somewhere. I believe the reason we set the weight to an i32 in Unleash-types is because we trust the rigor of our data validation in the typescript version as far as we can throw it and it's better to parse the wrong type and cast it than it is to fail the parse even though you have the data you need

unleash-yggdrasil/src/lib.rs
sighphyre: Your Google-Fu is most excellent. Right let's unpack this beast. 

In our function signature we declare `fn resolve_variant<'a>` which tells Rust that this function is going to receive some borrowed data and that the return value is _also_ going to be a borrow but it'll be borrowed against the incoming data. The 'a here is telling Rust that we're going to call this lifetime `'a`.

In the parameters we declare `variants: &'a Vec<CompiledVariant>` which tells Rust that this parameter has a lifetime of `'a` - this is the parameter that fulfils the lifetime declaration in the function signature and tells Rust that the lifetime of `'a` is as long as variants lives.

In the return value we have `Option<&'a CompiledVariant>` so now we're telling Rust that the return value must live at least as long as whatever data is bound to the `'a` lifetime.

So in English: we're borrowing a chunk of data called variants, when we return the data from this function, the data in variants can't be discarded before the return value is.

nunogois: That's perfect and really helpful, thanks Simon! 
@gastonfournier you may find this insightful as well.

PR #33, opened 2023-04-18
task: Bump unleash types to 0.10.0

chriswk:
Having updated unleash types to 0.10.0 to handle properties as an object map when using it as a query parameter we need yggdrasil to be on the same version.


gastonfournier (approved):
LGTM!

sighphyre (approved):
LGTM!

PR #34, opened 2023-04-26
fix: fix variant stickiness

teqm:
## About the changes

I noticed that `.get_variant()` doesn't work properly - the logic behind stickiness didn't consider `default` as a potential value (which I believe is present for every variant by "default"?). This caused `.resolve_variant()` to return `None`, which resulted in the default disabled variant being returned for the `.get_variant()` calls. 

While trying to fix that I refactored the variant stickiness logic to match the behavior of other SDKs (e.g. node, go) - unless that was intentional? :thinking:

Also fallback random stickiness was using `[0, 100)` range and the total variants weight sums up to 1000, so I believe for it to work properly it should be `[0, total_weight)` instead, right?

### DIFF
--- a/unleash-yggdrasil/src/error.rs
+++ b/unleash-yggdrasil/src/error.rs
@@ -1,3 +1,0 @@
-pub enum YggdrasilError {
-    StickinessExpectedButNotFound,
-}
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
Oh nice catch!

> Also fallback random stickiness was using [0, 100) range and the total variants weight sums up to 1000, so I believe for it to work properly it should be [0, total_weight) instead, right?

100% correct, completely missed this, it changed while we were building this project and we didn't update it! Thanks for the fix!

PR #36, opened 2023-05-16
fix: redesigns the way stickiness is handled in the grammar to support default and random

sighphyre:
## What

This completely reworks the way that the `default` and `random` keywords are handled by the grammar so they work correctly and remove a bunch of ugly edge cases in the `rollout_constraint`. As a side effect, this allows us to specify much more flexible rules for stickiness.

## Why

Firstly because `random` was completely ignored in the strategy upgrade path and that needed to be fixed. This caused the strategy to instead try to use a custom context field called "random" as its stickiness property, which short circuited the entire strategy to false. 

However, the bulk of the changes here are to fix the approach that led to the broken code in the first place. This was caused by incorrectly treating the stickiness property as if it were a special case of a context property. That doesn't model a stickiness property correctly, since they're more like a context property unless they're one of two keys words - "default" or "random". I've taken a very different approach here for each keyword:

"random - This has been handled in the strategy upgrade path correctly now. It's treated as a special case when resolving a context property in the same way that the built in context properties like "userId" and "sessionId" are.  

"default" - This is now expressed as a new type in the grammar as  chain of null coalescing context properties, which models how stickiness works correctly without edge case handling. This removes a lot of brittle code when actually executing the strategy

### DIFF
--- a/unleash-yggdrasil/src/strategy_grammar.pest
+++ b/unleash-yggdrasil/src/strategy_grammar.pest
@@ -81,10 +81,8 @@ constraint = {
     numeric_constraint = { context_value ~ ordinal_operation ~ num }
     semver_constraint = { context_value ~ ordinal_operation ~ semver }
     rollout_constraint = { percentage ~ stickiness_param? ~ group_id_param?}
-        stickiness_param = { "sticky on " ~ stickiness_option ~ ( NULL_COALESCE ~ stickiness_option)* }
+        stickiness_param = { "sticky on " ~ context_value ~ ( NULL_COALESCE ~ context_value)* }
         group_id_param = { "with group_id of" ~ string }
-
-stickiness_option = { context_value ~ ( NULL_COALESCE ~ context_value)* }
 
 context_value = { random | user_id | session_id | remote_address | app_name | environment | current_time | property }
     user_id         = { "user_id" }
--- a/unleash-yggdrasil/src/strategy_grammar.pest
+++ b/unleash-yggdrasil/src/strategy_grammar.pest
@@ -29,7 +29,9 @@ string_list = { "[" ~ string ~ ( "," ~ string )* ~ "]" }
 semver_list = { "[" ~ semver ~ ( "," ~ semver )* ~ "]" }
 
 WHITESPACE = _{ " " | "\t" }
+NULL_COALESCE = _{ "|" }
 
+random = { "random" }
 
 boolean_operation = _{ and | or }
     and     = { "and" }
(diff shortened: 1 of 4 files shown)


unleash-yggdrasil/src/strategy_grammar.pest
```diff
@@ -79,17 +81,16 @@ constraint = {
     numeric_constraint = { context_value ~ ordinal_operation ~ num }
     semver_constraint = { context_value ~ ordinal_operation ~ semver }
     rollout_constraint = { percentage ~ stickiness_param? ~ group_id_param?}
-        stickiness_param = { "sticky on " ~ context_value }
+        stickiness_param = { "sticky on " ~ context_value ~ ( NULL_COALESCE ~ context_value)* }
```
sighphyre, line 84: This is the only important change here (everything else is QoL), this allows us to express things like `sticky on user_id | session_id | random`

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -98,11 +98,18 @@ fn context_value(mut node: Pairs<Rule>) -> ContextResolver {
     }
 }
 
-// This is a special property, it's functionally identical to 'context_value'
```
sighphyre, line 101: Well this comment turned out to be spectacularly wrong

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -98,11 +98,18 @@ fn context_value(mut node: Pairs<Rule>) -> ContextResolver {
...
-// and our rule here does nothing but call 'context_value' to resolve the value
-// this only gets a special place so that our syntax in the grammar is nice
-fn stickiness_param(mut node: Pairs<Rule>) -> ContextResolver {
-    context_value(node.next().unwrap().into_inner())
+pub(crate) fn coalesce_context_property(node: Pairs<Rule>) -> ContextResolver {
```
sighphyre, line 101: For the unrusted, this function parses our context values into `(context) => Option<String>` functions and puts them in 'stickiness_resolvers' then moves them into a closure that executes them one by one, returning the results of the first one that returns a non null value

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -670,14 +653,6 @@ mod tests {
         assert_eq!(rule(&context), expected);
     }
 
-    #[test_case("55% with group_id of \"Feature.flexibleRollout.userId.55\"", true)]
```
sighphyre, line 673: No longer supported, all paths in the strategy upgrade now force a stickiness

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -274,39 +281,15 @@ fn semver_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
 fn rollout_constraint(mut node: Pairs<Rule>) -> RuleFragment {
     let percent_rollout = percentage(node.next().unwrap());
 
-    let second = node.next();
-    let third = node.next();
-
-    let (stickiness_getter, group_id) = match (second, third) {
```
sighphyre, line 280: Edge cases yeeted :)

gastonfournier (approved):
This looks good!

PR #37, opened 2023-06-02
fix: support for arbitrary strings

sighphyre:
Adds support for arbitrary strings in the grammar by allowing atom escaping of quotation marks within the string rule. For bonus points, we also include escaping of those characters.

### DIFF
--- a/unleash-yggdrasil/src/strategy_grammar.pest
+++ b/unleash-yggdrasil/src/strategy_grammar.pest
@@ -3,9 +3,7 @@ num = @{ int ~ ("." ~ ASCII_DIGIT*)? ~ (^"e" ~ int)? }
 
 percentage = { ("100" | (ASCII_DIGIT ~ ASCII_DIGIT?)) ~ "%" }
 
-string = @{ "\"" ~ ( unicode_characters | punctuation_characters )* ~ "\"" }
-    unicode_characters = { LETTER | MARK | NUMBER | SYMBOL | SEPARATOR | DASH_PUNCTUATION }
-    punctuation_characters = { "." | "&" | "/" | "_" | "@" | "!" }
+string = @{ "\"" ~ ( !("\\\"" | "\"") ~ ANY | "\\" ~ "\"")* ~ "\"" }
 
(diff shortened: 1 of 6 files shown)


gastonfournier (approved):
LGTM!

sighphyre:
Moving this to a draft temporarily. This should fix this exact issue but I'd like to get some prop tests in here to catch other things I may have missed

unleash-yggdrasil/src/strategy_grammar.pest
```diff
@@ -3,9 +3,7 @@ num = @{ int ~ ("." ~ ASCII_DIGIT*)? ~ (^"e" ~ int)? }
 
 percentage = { ("100" | (ASCII_DIGIT ~ ASCII_DIGIT?)) ~ "%" }
 
-string = @{ "\"" ~ ( unicode_characters | punctuation_characters )* ~ "\"" }
-    unicode_characters = { LETTER | MARK | NUMBER | SYMBOL | SEPARATOR | DASH_PUNCTUATION }
-    punctuation_characters = { "." | "&" | "/" | "_" | "@" }
+string = @{ "\"" ~ ( !("\\\"" | "\"") ~ ANY | "\\" ~ "\"")* ~ "\"" }
```
sighphyre, line 6: Reads loosely as match anything that starts with a quote but allow escaping quotes through \" characters. This should tolerate more or less anything unicodey and correctly escape the other grammar fragments

unleash-yggdrasil/tests/grammar_prop_tests.rs
```diff
@@ -0,0 +1,12 @@
+use proptest::prelude::*;
+use unleash_yggdrasil::strategy_parsing::compile_rule;
+
+proptest! {
+    #[test]
+    fn test_compile_rule(input in any::<String>().prop_filter("Exclude strings with \\\", \", \\ and empty strings", |s| !s.is_empty() && !s.contains("\\\"") && !s.contains("\"") && !s.contains("\\"))) {
```
sighphyre, line 6: We just strip out cases where we have our escaping characters. In most cases those will fail by design, in the cases where they're appropriate they're covered by concrete tests elsewhere

gastonfournier (approved):
LG! After second review!

PR #38, opened 2023-06-20
Language bindings

sighphyre:
## Introduction

This pull request is a significant step towards integrating the Yggdrasil engine across multiple programming languages via Foreign Function Interface (FFI) wrappers. These wrappers create an interface between the high-level programming languages and the low-level Yggdrasil library written in Rust.

The pull request includes FFI wrappers for the following languages: Go, Ruby, Python, Java, .NET (C#), and PHP. It essentially builds a foundational layer for these languages to utilize the functionalities of Yggdrasil.

### How to Read this PR
To get the most out of this pull request, it is suggested to approach it in the following order:

**Understand the Rust FFI Code**: Start by reading and understanding the Rust FFI implementation in `yggdrasilffi`. This is the low level code that exposes the native code for consumption by the host languages. This code assigns memory on the heap and returns pointers to the host code for that memory. As such, you're going to see a lot of unsafe code here by design, but it's crucial to ensure safety in memory handling to avoid issues such as memory leaks or double frees. The intent is that whenever Rust assigns memory and returns a pointer, the host language must be responsible for returning that pointer back to the Rust code so that it can be correctly deallocated. This code should never panic, since unwinding a panic across an FFI boundary is undefined behavior

**Choose a Familiar Language**: After gaining a good understanding of the Rust FFI, proceed to the wrapper implementation of a language you are comfortable with. This will help you understand how the FFI functions are used and how the interaction between the high-level language and the low-level Rust library occurs. It's worth reading the README to understand how to build and test the code.

**Review Other Languages**: Once you're comfortable that you understand how the FFI library is used in a language you're comfortable with, the other languages follow exactly the same patterns.

### Decisions and Assumptions

- This provides a single unified FFI wrapper for all the host languages as opposed to custom FFI layers. It's possible to build custom FFI wrappers for each of our SDK languages but this gives us the ability to battle test the implementation across multiple languages and ensure that this approach works for arbitrary languages

- JSON is the chosen communication protocol. While it's possible to pass pointers across the FFI boundary, this gets unwieldy very rapidly and the scope of leaks, double frees and use after frees becomes high. The JSON approach was chosen to reduce potential issues with memory handling. However, if this becomes problematic for performance reasons, we can rewrite individual layers while leaning on a working set of tests to ensure safety and correctness.

- Leaning on the Client Spec tests is okay. This work is more tested than it appears to be. Yggdrasil domain logic itself is extensively tested with its own test suites. Testing the Rust FFI layer directly doesn't make a lot of sense because the real things that we're concerned about - memory handling, can't be directly exercised without consuming that layer. As a result, each of the implementations has a full client spec suite. This is less to assert that the domain is respected and more to assert that data crosses the FFI boundary without crashes. 

### Future Work
While this is a substantial move towards enabling multi-language support for Yggdrasil, there are certain limitations and scope boundaries of the current work:

**Memory Management**: The current implementation discards the metrics coming back from the FFI layer, because this is heap allocated by the FFI layer, this leaks memory. We also need to build some tests on top of something like Valgrind to check that there aren't accidental leaks elsewhere.

**Package Management**: Part of the package management solution needs to address the way we're consuming the FFI library. Right now this is setup purely for development, this will not work for production.

**Metrics**: As of now, metrics functionality is not supported in these FFI wrappers. Adding metrics capabilities is an important aspect and will be part of future work.

The limitations mentioned above are not hindrances but rather, they outline the direction of future work in enhancing the usability and efficiency of these libraries.


unleash-yggdrasil/src/strategy_grammar.pest
```diff
@@ -10,7 +10,7 @@ semver = @{ ASCII_DIGIT+ ~ "."  ~ ASCII_DIGIT+ ~ "."  ~ ASCII_DIGIT+ ~ semver_pa
     semver_build    = @{ "+" ~ semver_fragment }
         semver_fragment = @{ (ASCII_ALPHANUMERIC | "-")+ ~ ("." ~ (ASCII_ALPHANUMERIC | "-")+)* }
 
-date = @{ quad_fragment ~ "-" ~ pair_fragment ~ "-" ~ pair_fragment ~ "T" ~ pair_fragment ~ ":" ~ pair_fragment ~ ":" ~ pair_fragment ~ "." ~ trip_fragment ~ timezone_fragment }
+date = @{ quad_fragment ~ "-" ~ pair_fragment ~ "-" ~ pair_fragment ~ "T" ~ pair_fragment ~ ":" ~ pair_fragment ~ ":" ~ pair_fragment ~ ("." ~ trip_fragment)? ~ timezone_fragment }
```
sighphyre, line 13: Dotnet does a thing where it truncates milliseconds when parsing dates. This isn't an existing bug in Yggdrasil since that's controlled by Yggdrasil's own date handling, but when you're using Dotnet, this doesn't parse. This allows us to use dates without millisecond precision

java-engine/lib/src/main/java/unleash/engine/Context.java
```diff
@@ -0,0 +1,69 @@
+package unleash.engine;
```
chriswk, line 1: We typically use the namespace `io.getunleash`, so this could/should probably be `io.getunleash.engine`

java-engine/lib/src/main/java/unleash/engine/UnleashEngine.java
```diff
@@ -0,0 +1,62 @@
...
+        UnleashFFI.INSTANCE.engine_free(ptr);
+    }
+
+    public void takeState(String toggles) {
+        UnleashFFI.INSTANCE.engine_take_state(ptr, toggles);
+    }
+
+    public boolean isEnabled(String name, Context context) throws JsonProcessingException {
+        String jsonContext = mapper.writeValueAsString(context);
+        return UnleashFFI.INSTANCE.engine_is_enabled(ptr, name, jsonContext);
```
chriswk, line 52: I'm afraid this will both be rather slow, as well as cause allocations of Strings, though they should never enter old gen.

go-engine/unleash_engine.go
```diff
@@ -0,0 +1,111 @@
...
+		Environment:   environment,
+		AppName:       appName,
+		CurrentTime:   currentTime,
+		RemoteAddress: remoteAddress,
+		Properties:    properties,
+	}
+}
+
+func main() {
+
```
nunogois, line 108:
```suggestion

```

go-engine/unleash_engine.go
```diff
@@ -0,0 +1,111 @@
...
+	ptr := C.engine_new()
+	fmt.Println("Hi", ptr)
```
nunogois, line 110: I'm guessing this is here for debug purposes, maybe it should be cleaned up.
```suggestion

```

java-engine/README.md
```diff
@@ -0,0 +1,23 @@
+# Java Bindings to Yggdrasil
+
+## Build
+
+We we gradle here:
```
nunogois, line 5: Typo? 😄

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -0,0 +1,103 @@
...
+        this.state = engine_new();
+    }
+
+    public void Dispose()
+    {
+        engine_free(this.state);
+        GC.SuppressFinalize(this);
+    }
+
+    public void TakeState(string json)
```
daveleek, line 72: So this is how you prep the engine behind the scenes, when you've fetched the togglestate from Unleash you feed it through here and then you can keep invoking from IsEnabled etc?

How is this state kept? Once per instance? Or is it per thread/process?

nunogois (approved):
This is without a doubt one of the coolest projects I've seen.
Even though some of this stuff may have a flown a bit over my head, I think you did a fantastic job with it.

go-engine/unleash_engine.go
sighphyre: I can always rely on you to find my silly debugs that I leave in hahaha, I need a linter for this

java-engine/README.md
sighphyre: We do do gradle here, don't see the problem

(Nice catch, absolute typo, fixed now!)

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    [DllImport(DLL_PATH)]
+    private static extern IntPtr engine_get_variant(IntPtr ptr, string toggle_name, string context);
+
+    [DllImport(DLL_PATH)]
+    private static extern void engine_free_variant_def(IntPtr ptr);
+
+    public UnleashEngine()
+    {
+        this.state = engine_new();
```
daveleek, line 63: Ah so this is what keeps a reference to the rust engine-object?

dotnet-engine/UnleashEngine/UnleashEngine.cs
sighphyre: This is a great question! Unfortunately the answer is complex and it requires some understanding of how the memory is allocated.

State is handled Rust side and the host language gets a pointer to that state. That's allocated by Rust's allocator, not the .NET allocator. So state is once per pointer returned by the constructor in the Rust code. In practice that's as many times as you want per thread (singleton stuff needs to be related to the toggle fetcher not this code) and memory between Rust and host language is handled by OS memory mapper so processes are isolated but in sync across Rust and host language

java-engine/lib/src/main/java/unleash/engine/Context.java
sighphyre: Great suggestion, thanks! Fixed

java-engine/lib/src/main/java/unleash/engine/UnleashEngine.java
sighphyre: I'm more worried about the overhead of serialization than I am about string allocation here but this is a fair point. There's definitely more work that needs to be done here but this needs benchmarks before changes are made. 

There's a few things we can do here, in order of increasing complexity:
- Pass byte buffers instead of strings, read that on the Rust side into a context struct
- Pass pointers to byte buffers instead of strings, deref the pointer and read that into a context
- Pass a pointer array to each item in the context, build the struct up piece by piece in Rust

Golang hates each and every one of these approaches because the FFI layer in Golang was designed by a turnip and so for the POC I've chosen a standard approach that works for anything even if it's not blazingly fast. This isn't unusual in high level FFI wrapping libraries when passing complex datatypes but we can probably do better.

We also need to make a decision around how fast this needs to be and balance that against the complexity and risk of dipping into complex pointer maths

dotnet-engine/UnleashEngine/UnleashEngine.cs
sighphyre: This is what keeps a reference to a _pointer_ to an engine object. This is important because if you lose this pointer without calling the function responsible for deallocating the memory bound to the pointer in Rust, you'll get a memory leak. Likewise, calling this multiple times will trigger a use after free. It's up to the caller to behave like an adult

unleash-yggdrasil/src/strategy_grammar.pest
sighphyre: Haven't we all? xD

Never change, .NET, never change

yggdrasilffi/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+    str::Utf8Error,
+};
+
+use unleash_types::{client_features::ClientFeatures, client_metrics::MetricBucket};
+use unleash_yggdrasil::{Context, EngineState};
+
+#[derive(Debug)]
+enum FFIError {
+    Utf8Error,
+    NulError,
```
thomasheartman, line 12: Just a curiosity question: why is this `NulError` as opposed to `NullError` (or maybe `NilError`)? Does it map to something?

chriswk (approved):
The Rust code looks good from my experience with it. 
The Java code and the deserializing objects to String to pass it across the FFI boundary worries me a bit, but as an initial effort to get something working it's a lot better than what I feared.

It'll be exciting to see where we head with this the next quarter.

yggdrasilffi/cbindgen.toml
thomasheartman, line 1: Is this file empty on purpose? Or was it added by mistake?

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -0,0 +1,103 @@
...
+        engine_take_state(state, json);
+    }
+
+    public bool IsEnabled(string toggleName, Context context)
+    {
+        string contextJson = JsonSerializer.Serialize(context, options);
+        return engine_is_enabled(state, toggleName, contextJson);
+    }
+
+    public Variant? GetVariant(string toggleName, Context context)
```
thomasheartman, line 83: So this should only ever return `null` if the variantPtr is invalid, right? Is there a way to get around that? I'm guessing no? And if I understand correctly, this is intended to be the public API, yeah? Just seems like that would be a change from how we do it today. Would it make sense to return the disabled variant in that case? 🤔

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -0,0 +1,103 @@
+﻿
+using System.Runtime.InteropServices;
+using System.Text.Json;
+
+
+namespace Unleash;
+
+public class Context
```
thomasheartman, line 8: Some of these properties are required in the current implementation. Do we want to make them all optional now?

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+    public string? UserId { get; set; }
+    public string? SessionId { get; set; }
+    public string? RemoteAddress { get; set; }
+    public string? Environment { get; set; }
```
thomasheartman, line 13: Question for discussion: can/should we remove this property? Is yggdrasil a good chance to finally lose this deprecated property that is causing a lot of confusion?

java-engine/lib/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * This Java source file was generated by the Gradle 'init' task.
```
thomasheartman, line 2: Is this still true?

thomasheartman (approved):
Yeah, this is super duper cool 🏆 I've had a look through to the best of my ability, and it looks great! I've left some questions, but it's mostly for curiosity. No real change requests here, I think. Well done! 👏🏼 👏🏼 👏🏼

yggdrasilffi/cbindgen.toml
sighphyre: Yuh, cbindgen wants it to run correctly but it's okay for it to be empty

dotnet-engine/UnleashEngine/UnleashEngine.cs
sighphyre: Yeah good spot, this needs a bit of love here, I think it's okay for a POC but this should eventually return a non nullable type

And yeah, this will only happen if the caller does something very, very dumb like passing an invalid pointer. The null pointer from Rust is correct here in that there's little that the Rust code can do and attempting to proceed is fundamentally broken and the caller needs to know about it, but the wrapping code should handle this more elegantly

dotnet-engine/UnleashEngine/UnleashEngine.cs
sighphyre: I think I have the dumb, not sure what you mean here? The implementation in .NET, Yggdrasil or our existing standard expectations of how an SDK works?

dotnet-engine/UnleashEngine/UnleashEngine.cs
sighphyre: Oh you sharp eyed ninja. That's a fantastic idea! Yes I believe this can come out now, it'll need a little fiddling in a separate PR though

java-engine/lib/src/main/java/io/getunleash/engine/UnleashEngine.java
sighphyre: Good catch! It was generated by Gradle but I fiddled, taken it out, thank you!

ruby-engine/unleash_engine_ffi.rb
```diff
@@ -0,0 +1,49 @@
...
+  attach_function :engine_is_enabled, [:pointer, :string, :string], :bool
+  attach_function :engine_get_variant, [:pointer, :string, :string], :pointer
+  attach_function :engine_free_variant_def, [:pointer], :void
+
+  def initialize
+    @engine_state = UnleashEngine.engine_new
+    ObjectSpace.define_finalizer(self, self.class.finalize(@engine_state))
+  end
+
+  def self.finalize(engine_state)
```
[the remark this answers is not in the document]
sighphyre: This is another great question!

Short answer - because Ruby is weird

Long answer - we want to generate a lazy callback to clean up the engine state pointer and deallocate the memory. This can't be a object method because then we're generating a reference to the very object we're trying to clean up, which can cause the Ruby VM to leak the memory it allocated in its own memory space. So this boils down to the way objects are garbage collected in Ruby

dotnet-engine/UnleashEngine/UnleashEngine.cs
sighphyre: Ah! Gotcha! That makes sense. I don't believe this is an issue for Yggdrasil. Yggdrasil doesn't really care, it'll evaluate your context whether you set these or not. Enforcing appName has more to do with how metrics are sent and how backupfiles are saved in the SDKs. So if the caller needs these it needs to set them. It's entirely reasonable for the caller to require appName to do other things but not require it for evaluating a toggle

thomasheartman: Right! I think this connects to my other question and shows that I didn't really understand how this to be managed. I was under the impression that these wrappers would take the place of the SDKs. In that case we're all good. Thanks for explaining 🙏🏼

sighphyre: Ah, now the question makes a lot of sense.

Yep, this isn't designed to take over our SDKs. We have, for the most part, really robust and battle tested layers that do things like disk caching and http in our existing SDKs. This work is to take over the domain logic from the SDKs, not do anything else. So my intent is to eat the domain of logic of SDKs from the inside out with this - whatever the inverse of a strangler pattern is

ruby-engine/unleash_engine_ffi.rb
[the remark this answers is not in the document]
rarruda: In this case it makes the method static. So it allows for what @sighphyre  is saying.

PR #39, opened 2023-06-23
feat: add single toggle resolve method

sighphyre:
Adds a resolve method for Yggdrasil that can compute a single toggle. This better supports Edge, which currently needs to iterate the whole hashmap to resolve a toggle


gastonfournier (approved):
LGTM!

PR #41, opened 2023-07-14
feat: add support for strategy variants

sighphyre:
This adds support for strategy variants, as defined here:  https://github.com/Unleash/client-specification/blob/main/specifications/16-strategy-variants.json 

Metrics will not work against the strategy variants but they're not in use anywhere right now

This works a little differently from how this is likely going to be implemented in other SDKs, since it taps into the Yggdrasil grammar. The flow works like this:

- Compile the strategy to a Vec of Yggdrasil rule fragments, granular to the strategy level
- When a variant is requested, find the first strategy that resolves to true and get the strategy variants for that, reuse the strategy stickiness for that variant stickiness
- If the above is Some and not None, then pass that variant set to the standard variant resolution flow
- Else check for a standard variant

Please excuse the noise, Clippy woke up and chose violence this morning

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -979,45 +979,6 @@ mod test {
     }
 
     #[test]
-    pub fn strategy_variants_fallback_to_disabled_variant() {
-        let mut compiled_state = HashMap::new();
-        compiled_state.insert(
-            "some-toggle".to_string(),
-            CompiledToggle {
-                name: "some-toggle".into(),
-                enabled: true,
-                compiled_strategy: Box::new(|_| true),
-                variants: vec![CompiledVariant {
-                    name: "should-be-ignored".into(),
-                    weight: 100,
-                    stickiness: None,
-                    payload: None,
-                    overrides: None,
-                    count: AtomicU32::new(0),
-                }],
-                compiled_variant_strategy: Some(vec![(
-                    Box::new(|_| true),
-                    vec![CompiledVariant {
-                        count: AtomicU32::new(0),
-                        name: "don't-ignore-me".into(),
-                        weight: 100,
-                        stickiness: None,
-                        payload: None,
-                        overrides: None,
-                    }],
-                )]),
-                ..CompiledToggle::default()
-            },
-        );
-        let state = EngineState {
-            compiled_state: Some(compiled_state),
-            ..Default::default()
-        };
-        let variant = state.get_variant("some-toggle", &Context::default());
-        assert_eq!(variant.name, "don't-ignore-me".to_string());
-    }
-
-    #[test]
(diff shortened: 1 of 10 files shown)


unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -18,6 +18,28 @@ pub fn upgrade(strategies: &Vec<Strategy>, segment_map: &HashMap<i32, Segment>)
     rule_text
 }
 
+pub fn build_variant_rules(
```
sighphyre, line 21: This is the magic that makes this work - we build a set of rules for each individual strategy and return a tuple of [rule, Vec<Variant>]

unleash-yggdrasil/src/lib.rs
```diff
@@ -89,16 +93,49 @@ fn build_segment_map(segments: &Option<Vec<Segment>>) -> HashMap<i32, Segment> {
...
+                    (
+                        rule,
+                        strategy_variants
+                            .iter()
+                            .map(|strategy_variant| CompiledVariant {
+                                name: strategy_variant.name.clone(),
+                                weight: strategy_variant.weight,
+                                stickiness: Some(stickiness.clone()),
+                                payload: strategy_variant.payload.clone(),
+                                overrides: None,
```
gastonfournier, line 115: Just checking: are overrides not needed?

unleash-yggdrasil/src/strategy_upgrade.rs
gastonfournier: We're only doing this for strategies with variants https://github.com/Unleash/yggdrasil/pull/41/files#diff-fcd282bba31d2f9abf99d8187bc54b14f83a5863c9c6dcc9302204af1d7b8abbR27

I'm wondering what would happen if you have some strategies with variants and some without...

gastonfournier (commented):
Looks good, just trying to think of some edge cases

unleash-yggdrasil/src/strategy_upgrade.rs
sighphyre: Won't matter, all strategies will go through this flow, if they have variants, they'll get a rule for that, if they don't, they'll get none

When we check the variant, we first check if the strategy has a compiled variant set, if it does we'll use that, if it doesn't we'll fallback to the "normal" variant or none if those also don't exist

unleash-yggdrasil/src/lib.rs
sighphyre: Yep,  they're specifically excluded in the spec. unleash-types specifies a special strategy variant which holds only the things that are required. We're mapping this to a base variant here so we can use the existing variant flow and intentionally ignoring the override here

gastonfournier (approved):
Thanks @sighphyre for the explanations, just confirmed my thoughts! LGTM"

PR #42, opened 2023-08-14
refactor: better public api for ffi consumers

sighphyre:
This makes two major changes to Yggdrasil:

1) Redesigns metrics a little to separate them from the actual compiled toggles. This is mostly for convenience - `CompiledToggle` is arbitrarily clonable, whereas we use Atomic integers for counting metrics, which are not clonable. This allows us to better track toggles and variants that are not defined, without hacking in fake toggles into the cache

2) Exposes a few new lower level constructs to the public API - `check_toggle`, `check_variant`, `count_toggle`, `count_variant` and `get_metrics`. `check_toggle` and `check_variant` are for accessing the toggle state in a way that the caller can determine whether or not the toggle actually exists, which is a hard requirement for some of our SDKs. `count_toggle` and `count_variant` are to allow the clients to track metrics in the ways that they want to. I want to say `get_metrics` is obvious

### DIFF
--- a/unleash-yggdrasil/Cargo.toml
+++ b/unleash-yggdrasil/Cargo.toml
@@ -19,6 +19,7 @@ semver = "1.0.17"
 convert_case = "0.6.0"
 unleash-types = "0.10.4"
 chrono = "0.4.26"
+dashmap = "5.5.0"
 
 [dependencies.serde]
 features = ["derive"]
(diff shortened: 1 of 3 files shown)


gardleopard:
What is ffi?

sighphyre:
> What is ffi?

Foreign Function Interface. To brutally summarise a complex thing - it's how we call Rust/C/C++ from other languages

gardleopard (approved):
LGTM

PR #43, opened 2023-09-18
feat: make ruby useful

sighphyre:
THIS IS A DRAFT. It's going to look a little (lot) ugly while it gets iterated on 

This is an attempt to bring Yggdrasil up to the point where the Ruby SDK can use it without making too many sacrifices within its own code.


nunogois (approved):
LGTM

PR #44, opened 2023-09-20
feat: add support for ignore_case for in and not_in operators

gastonfournier:
- Add support to 2 new operators following the same convention as other operators (`in_ignore_case` and `not_in_ignore_case`)
- Add test cases for different scenarios

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Yggdrasil is a Rust project designed to create the core of the Unleash SDK domai
 
 ## Building the Core
 
-Easy enough - run `cargo build` from the root of the project. You'll need an up to date set of Rust tools to do this.
+Easy enough - run `cargo build` from the root of the project. You'll need an up to date set of Rust tools to do this (`rustup update stable`).
 
 To run the client specs, you'll first need to clone them:
 
@@ -42,4 +42,5 @@ Start by setting up and activating a virtual environment in the python-sdk folde
 python3 -m venv venv
 source venv/bin/activate
 ```
+
 Install [maturin](https://github.com/PyO3/maturin) by executing `pip install maturin` in your shell. Then you can run `maturin develop`.
--- a/unleash-yggdrasil/src/strategy_grammar.pest
+++ b/unleash-yggdrasil/src/strategy_grammar.pest
@@ -56,7 +56,7 @@ string_list_operation_without_case = {
     ends_with_ignore_case       = { "ends_with_any_ignore_case" }
     contains_any_ignore_case    = { "contains_any_ignore_case" }
 
-list_operation = { "in" | "not_in" }
+list_operation = { "in" | "not_in" | "in_ignore_case" | "not_in_ignore_case" }
 
 invert_operation = { "!" }
 
--- a/unleash-yggdrasil/src/strategy_grammar.pest
+++ b/unleash-yggdrasil/src/strategy_grammar.pest
@@ -56,7 +56,7 @@ string_list_operation_without_case = {
     ends_with_ignore_case       = { "ends_with_any_ignore_case" }
     contains_any_ignore_case    = { "contains_any_ignore_case" }
 
-list_operation = { "in" | "not_in" | "in_ignore_case" | "not_in_ignore_case" }
+list_operation = { "in_ignore_case" | "not_in_ignore_case" | "in" | "not_in" }
 
 invert_operation = { "!" }
 
--- a/unleash-yggdrasil/src/strategy_parsing.rs
+++ b/unleash-yggdrasil/src/strategy_parsing.rs
@@ -125,7 +125,11 @@ fn context_property(mut node: Pairs<Rule>) -> ContextResolver {
         }
     })
 }
+// ! in  ["a", "b"]
+// not_in  ["a", "b"]
 
+// ! not_in  ["a", "b"]
+// in  ["a", "b"]
 fn to_ordinal_comparator(node: Pair<Rule>) -> OrdinalComparator {
     match node.as_str() {
         "<" => OrdinalComparator::Lt,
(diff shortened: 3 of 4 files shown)


unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -353,14 +357,30 @@ fn list_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
...
+
+                    ContentComparator::NotInIgnoreCase => match context_value {
+                        Some(context_value) => {
+                            let needle = context_value.to_lowercase();
+                            values.iter().find(|x| x.to_lowercase() == needle).is_none()
+                        }
                         None => true,
                     },
                 }
+                .invert(inverted)
```
gastonfournier, line 383: Not sure if this is standard in rust, or if it's better to keep everything together. Lines started to grow big and this little change made it more readable, but I'm ok moving it back (also for consistency with the other operations)

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -353,14 +357,30 @@ fn list_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
...
                     },
                     ContentComparator::NotIn => match context_value {
-                        Some(context_value) => !values.contains(&context_value).invert(inverted),
+                        Some(context_value) => !values.contains(&context_value),
+                        None => true,
+                    },
+                    ContentComparator::InIgnoreCase => match context_value {
+                        Some(context_value) => {
+                            let needle = context_value.to_lowercase();
+                            values.iter().find(|x| x.to_lowercase() == needle).is_some()
```
sighphyre, line 370: Thank you Clippy! This can be written as

```suggestion
       values.iter().any(|x| x.to_lowercase() == needle)
```

Which avoids the `is_some()` check

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -353,14 +357,30 @@ fn list_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
...
+                        }
+                        None => false,
+                    },
+
+                    ContentComparator::NotInIgnoreCase => match context_value {
+                        Some(context_value) => {
+                            let needle = context_value.to_lowercase();
+                            values.iter().find(|x| x.to_lowercase() == needle).is_none()
```
sighphyre, line 378: Same as above

```suggestion
               !values.iter().any(|x| x.to_lowercase() == needle)
```

unleash-yggdrasil/src/strategy_parsing.rs
sighphyre: Unfortunately that's a logic change here, since this will also flip the state when the context value doesn't exist. 

Annoying that neither the existing tests nor the client spec tests picked it up. 

This might be more correct but I think it needs a bit more thought than a 5 minute glance

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -738,6 +759,13 @@ mod tests {
     #[test_case("user_id contains_any_ignore_case [\"EMAIL\"]", true)]
     #[test_case("user_id ends_with_any_ignore_case [\".COM\"]", true)]
     #[test_case("user_id starts_with_any_ignore_case [\"SOME\"]", true)]
+    #[test_case("user_id in_ignore_case [\"some-EMAIL.com\"]", true)]
+    #[test_case("user_id in_ignore_case [\"noemail.com\",\"neither-THIS.com\"]", false)]
+    #[test_case("user_id not_in_ignore_case [\"notemail.com\"]", true)]
+    #[test_case(
+        "user_id not_in_ignore_case [\"notemail.com\",\"some-EMAIL.com\"]",
+        false
+    )]
```
sighphyre, line 768: Nice! These LGTM but I think we should have another set where the context value is also mixed case. These cover the case where

`"someuser" in_ignore_case ["sOMeUSer"]`

but it doesn't cover

`"sOMeUSer" in_ignore_case ["someuser"]`

unleash-yggdrasil/src/strategy_grammar.pest
```diff
@@ -56,7 +56,7 @@ string_list_operation_without_case = {
     ends_with_ignore_case       = { "ends_with_any_ignore_case" }
     contains_any_ignore_case    = { "contains_any_ignore_case" }
 
-list_operation = { "in" | "not_in" }
+list_operation = { "in_ignore_case" | "not_in_ignore_case" | "in" | "not_in" }
```
sighphyre, line 59: Didn't realise this when we spoke earlier but this might deserve a small redesign to the way the grammar handles these. This:

`userId in_ignore_case [1, 2, 4]`

Makes no sense and is valid right now

Up for pairing tomorrow if you want

README.md
```diff
@@ -8,7 +8,7 @@ Yggdrasil is a Rust project designed to create the core of the Unleash SDK domai
 
 ## Building the Core
 
-Easy enough - run `cargo build` from the root of the project. You'll need an up to date set of Rust tools to do this.
+Easy enough - run `cargo build` from the root of the project. You'll need an up to date set of Rust tools to do this (`rustup update stable`).
```
sighphyre, line 11: :heart:

sighphyre (commented):
Absolutely amazing first dip into Ygg! Don't want to merge it yet but I think you've got the core ideas down just fine!

unleash-yggdrasil/src/strategy_grammar.pest
gastonfournier: Nice suggestion, we fixed this

sighphyre (approved):
WHEE LGTM

sighphyre:
This isn't something we need in the grammar right now, so I'm going to let this hang for a bit. The point was a learning exercise

PR #46, opened 2023-10-04
fix: resolves the memory leaks in the ruby ffi layer

sighphyre:
This fixes the remaining memory leaks in Ruby, where we previously weren't freeing resources we explicitly needed to free. This also adds a script to the ruby engine folder that will exercise the engine with a large number of allocations and do some bare bones checks for obvious leaks 

Co-authored-by: nuno@getunleash.ai


nunogois (approved):
LGTM

PR #47, opened 2023-10-04
refactor: Yggdrasil FFI layer

nunogois:
Attempts to refactor Yggdrasil's FFI layer after reading through the code.

I'm a bit unsure about the method renames. I wanted simpler names on the interface, but also noticed that simpler names sometimes meant unexpected problems.

I tried to be a bit more explicit about the unsafes by making their scope a lot tighter (only unsafe what we need to unsafe). I also updated the docs for the methods and tried to make them consistent. Added some helper methods for a DRY-er approach. Includes some renames, moving some things around and some small typo corrections.

Tests are green, both `cargo test` at the top level and `rspec` in `ruby-engine`.

I hope this makes sense but I'm totally OK with simply discarding this if we don't see value in it 🙂

### DIFF
--- a/ruby-engine/lib/unleash_engine.rb
+++ b/ruby-engine/lib/unleash_engine.rb
@@ -36,39 +36,39 @@ class UnleashEngine
   extend FFI::Library
   ffi_lib File.expand_path(platform_specific_lib, __dir__)
 
-  attach_function :engine_new, [], :pointer
-  attach_function :engine_free, [:pointer], :void
+  attach_function :new_engine, [], :pointer
+  attach_function :free_engine, [:pointer], :void
 
-  attach_function :engine_take_state, %i[pointer string], :pointer
-  attach_function :engine_check_enabled, %i[pointer string string], :pointer
-  attach_function :engine_check_variant, %i[pointer string string], :pointer
-  attach_function :engine_get_metrics, [:pointer], :pointer
-  attach_function :engine_free_response_message, [:pointer], :void
+  attach_function :take_state, %i[pointer string], :pointer
+  attach_function :check_enabled, %i[pointer string string], :pointer
+  attach_function :check_variant, %i[pointer string string], :pointer
+  attach_function :get_metrics, [:pointer], :pointer
+  attach_function :free_response, [:pointer], :void
 
-  attach_function :engine_count_toggle, %i[pointer string bool], :void
-  attach_function :engine_count_variant, %i[pointer string string], :void
+  attach_function :count_toggle, %i[pointer string bool], :void
+  attach_function :count_variant, %i[pointer string string], :void
 
   def initialize
-    @engine_state = UnleashEngine.engine_new
-    ObjectSpace.define_finalizer(self, self.class.finalize(@engine_state))
+    @engine = UnleashEngine.new_engine
+    ObjectSpace.define_finalizer(self, self.class.finalize(@engine))
   end
 
-  def self.finalize(engine_state)
-    proc { UnleashEngine.engine_free(engine_state) }
+  def self.finalize(engine)
+    proc { UnleashEngine.free_engine(engine) }
   end
 
   def take_state(toggles)
-    response_ptr = UnleashEngine.engine_take_state(@engine_state, toggles)
+    response_ptr = UnleashEngine.take_state(@engine, toggles)
     take_toggles_response = JSON.parse(response_ptr.read_string, symbolize_names: true)
-    UnleashEngine.engine_free_response_message(response_ptr)
+    UnleashEngine.free_response(response_ptr)
   end
 
   def get_variant(name, context)
     context_json = (context || {}).to_json
 
-    variant_def_json_ptr = UnleashEngine.engine_check_variant(@engine_state, name, context_json)
+    variant_def_json_ptr = UnleashEngine.check_variant(@engine, name, context_json)
     variant_def_json = variant_def_json_ptr.read_string
-    UnleashEngine.engine_free_response_message(variant_def_json_ptr)
+    UnleashEngine.free_response(variant_def_json_ptr)
     variant_response = JSON.parse(variant_def_json, symbolize_names: true)
 
     return nil if variant_response[:status_code] == TOGGLE_MISSING_RESPONSE
@@ -78,9 +78,9 @@ class UnleashEngine
   def enabled?(toggle_name, context)
     context_json = (context || {}).to_json
 
-    response_ptr = UnleashEngine.engine_check_enabled(@engine_state, toggle_name, context_json)
+    response_ptr = UnleashEngine.check_enabled(@engine, toggle_name, context_json)
     response_json = response_ptr.read_string
-    UnleashEngine.engine_free_response_message(response_ptr)
+    UnleashEngine.free_response(response_ptr)
     response = JSON.parse(response_json, symbolize_names: true)
 
     raise "Error: #{response[:error_message]}" if response[:status_code] == ERROR_RESPONSE
@@ -89,17 +89,17 @@ class UnleashEngine
   end
 
   def count_toggle(toggle_name, enabled)
-    UnleashEngine.engine_count_toggle(@engine_state, toggle_name, enabled)
+    UnleashEngine.count_toggle(@engine, toggle_name, enabled)
   end
 
   def count_variant(toggle_name, variant_name)
-    UnleashEngine.engine_count_variant(@engine_state, toggle_name, variant_name)
+    UnleashEngine.count_variant(@engine, toggle_name, variant_name)
   end
 
   def get_metrics
-    metrics_ptr = UnleashEngine.engine_get_metrics(@engine_state)
+    metrics_ptr = UnleashEngine.get_metrics(@engine)
     metrics = JSON.parse(metrics_ptr.read_string, symbolize_names: true)
-    UnleashEngine.engine_free_response_message(metrics_ptr)
+    UnleashEngine.free_response(metrics_ptr)
     metrics[:value]
   end
 end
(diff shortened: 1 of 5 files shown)


ruby-engine/lib/unleash_engine.rb
```diff
@@ -36,37 +36,37 @@ class UnleashEngine
   extend FFI::Library
   ffi_lib File.expand_path(platform_specific_lib, __dir__)
 
-  attach_function :engine_new, [], :pointer
-  attach_function :engine_free, [:pointer], :void
+  attach_function :new_engine, [], :pointer
+  attach_function :free_engine, [:pointer], :void
 
-  attach_function :engine_take_state, %i[pointer string], :string, :pointer
```
sighphyre, line 42: Oh this is a lot better. This became smurf naming without me even realising it

yggdrasilffi/src/lib.rs
```diff
@@ -56,257 +63,241 @@ impl From<serde_json::Error> for FFIError {
     }
 }
 
-#[no_mangle]
-pub extern "C" fn engine_new() -> *mut c_void {
-    let state = EngineState::default();
-    Box::into_raw(Box::new(state)) as *mut c_void
+fn get_engine<'a>(engine_ptr: *mut c_void) -> Result<&'a mut EngineState, FFIError> {
```
sighphyre, line 66: I would rather we mark the function as unsafe itself, rather than the block of code that does the unsafe stuff. 

Putting the unsafe on the function level tells other programmers that "you need to know what you're doing here, if you pass the wrong thing or do the wrong thing with the return type, bad things may happen". In this case, the wrong thing can be passing a valid pointer to an invalid reference. 

Putting the unsafe inside the function without marking the function as unsafe tells other programmers "there might be some unsafe stuff in here, but don't worry, we did checking before and there's no way you, as the caller, can cause the bad things to happen"

sighphyre: (but to clear, this is a really nice cleanup)

yggdrasilffi/src/lib.rs
```diff
@@ -56,257 +63,241 @@ impl From<serde_json::Error> for FFIError {
...
+fn get_str<'a>(ptr: *const c_char) -> Result<&'a str, FFIError> {
+    if ptr.is_null() {
+        Err(FFIError::NullError("Null pointer".into()))
+    } else {
+        unsafe { CStr::from_ptr(ptr).to_str().map_err(FFIError::from) }
+    }
 }
 
-/// Frees the memory allocated for the engine
+fn get_json<T: DeserializeOwned>(json_ptr: *const c_char) -> Result<T, FFIError> {
```
sighphyre, line 82: Same as above - this is 100% an unsafe function and callers should know that

yggdrasilffi/src/lib.rs
```diff
@@ -56,257 +63,241 @@ impl From<serde_json::Error> for FFIError {
...
+/// These pointers should not be dropped for the lifetime of this function call.
+///
+/// This function must be called correctly in order to deallocate the memory allocated for the engine in
+/// the `new_engine` function. Failure to do so will result in a leak.
+#[no_mangle]
+pub unsafe extern "C" fn free_engine(engine_ptr: *mut c_void) {
+    if engine_ptr.is_null() {
         return;
     }
-    unsafe {
```
sighphyre, line 84: :+1:

yggdrasilffi/src/lib.rs
```diff
@@ -56,257 +63,241 @@ impl From<serde_json::Error> for FFIError {
...
+/// Null pointers will result in an error message being returned to the caller,
+/// but any invalid pointers will result in undefined behavior.
+/// These pointers should not be dropped for the lifetime of this function call.
 #[no_mangle]
-pub unsafe extern "C" fn engine_take_state(ptr: *mut c_void, json: *const c_char) -> *const c_char {
-    let response: Response<()> = take_state(ptr, json).into();
-    let json = serde_json::to_string(&response).unwrap();
-    CString::new(json).unwrap().into_raw()
-}
+pub extern "C" fn take_state(engine_ptr: *mut c_void, json_ptr: *const c_char) -> *const c_char {
```
sighphyre, line 133: Oh this is about as safe as rubbing yourself with honey and slapping a bear

Let's put the unsafe back here! Absolutely okay for you to isolate the unsafe code block to a small portion though, I think that pattern is very nice!

yggdrasilffi/src/lib.rs
```diff
@@ -56,257 +63,241 @@ impl From<serde_json::Error> for FFIError {
...
+    let result: Result<Option<()>, FFIError> = (|| {
```
sighphyre, line 134: Why the immediately executed closure? Is this so you can get the error handling with the single function? Not against it, just new to me as an idea

yggdrasilffi/src/lib.rs
```diff
@@ -56,257 +63,241 @@ impl From<serde_json::Error> for FFIError {
...
+/// but any invalid pointers will result in undefined behavior.
+/// These pointers should not be dropped for the lifetime of this function call.
+///
+/// This function must be called correctly in order to deallocate the memory allocated for the response in
+/// the `check_enabled`, `check_variant`, `count_toggle`, `count_variant` and `get_metrics` functions.
+/// Failure to do so will result in a leak.
 #[no_mangle]
-pub unsafe extern "C" fn engine_free_response_message(ptr: *mut c_char) {
-    if ptr.is_null() {
+pub unsafe extern "C" fn free_response(response_ptr: *mut c_char) {
```
sighphyre, line 214: :+1: 

Much nicer name!

sighphyre (commented):
This is pretty darn great actually. I think overall the refactor is a huge improvement. I think we should put the unsafe back though haha, trimming down on the unsafe blocks and removing it from the method signatures doesn't make things any safer, only more hidden :ninja: 

I 100% see value. I want it!

yggdrasilffi/src/lib.rs
nunogois: That's fair. Addressed in https://github.com/Unleash/yggdrasil/pull/47/commits/c7723e0b49e4a941e3dab9a39dcd39b3c3c64ded

My reasoning for this change was that it felt better to me to only specify `unsafe` when needed. Marking a method as `unsafe` means you can technically do anything unsafe inside that method and the compiler won't scream at you, which feels wrong. At that point, do we still mark the unsafe code inside the method as unsafe or not? Whatever we decide to do, the compiler won't care, which again, feels wrong to me.

However I understand that from an interface perspective it may make sense to mark the methods as unsafe.

yggdrasilffi/src/lib.rs
nunogois: Thanks, same as https://github.com/Unleash/yggdrasil/pull/47#discussion_r1348066338

yggdrasilffi/src/lib.rs
nunogois: 🍯🕺🐻
Loved the analogy 🤣 
Same as https://github.com/Unleash/yggdrasil/pull/47#discussion_r1348066338

yggdrasilffi/src/lib.rs
nunogois: Well, I wanted to standardize our methods, since they mostly consist of grabbing the values from the pointers we pass in and then returning a response based on the internal method we execute - Basically `Result<Option<T>, FFIError>`. With this in mind, we could have an auxiliary method like `result_to_json_ptr` that would convert our response to a pointer we could return. I also wanted to avoid having a separate auxiliary method for each public method to improve consistency, readability and maintainability.

So this is what I came up with. It feels clean to me but maybe it's actually more readable and maintainable to have separate methods? 🤔 I'm open to suggestions here!

nunogois:
Thank you for the review @sighphyre! I'm very happy you liked it.

I tried addressing the comments as best as I could. 

With these changes most methods return a response, so I also added a `free_response` in `unleash_engine.rb` to cover those cases. Perhaps we should consider adding a response to all methods, even if most are just OKs? My gut feeling is that it would provide a better and more consistent interface.

Feel free to do anything you want with this in the meantime: Change it, merge it, whatever makes sense to you.

Thanks!

yggdrasilffi/src/lib.rs
sighphyre: It does make a lot of sense from the interface perspective, unfortunately. Our methods here are definitely not safe and we can't make them safe, so we should be polite enough to tell other people's documentation/compilers that that's the case. Otherwise we run the risk of letting others call these methods blindly without understanding the consequences (or even that there are consequences).

What I was originally going for is something like

```fn unsafe some_public_method() { 
  let results = unsafe {
  // some stuff that may set fire to the universe
  }
  handle_stuff(results) //some safe method, where we can write polite code
}
```

But I don't know if that strategy worked out. It got a bit too messy.

I think what you have at this point is a pretty good balance. Kinda want to just merge it

yggdrasilffi/src/lib.rs
sighphyre: I can get behind this. There's a small cost to wrapping stuff in closures like this but given what the rest of the file is doing, it shouldn't be noticeable. It makes things nicer to read and cleaner

I'm okay with responses for everything except the two metrics methods. A caller can't respond to a failure there in any meaningful way so it makes sense for them to be returnless. Everything else I think now has a standard response format

sighphyre (approved):
LGTM, it's a nice cleanup! I think we have some work to do to think a little on how we use unsafe in this subproject but right now I think this is the cleanest I've seen this. I vote we merge and iterate if we get something smarter

yggdrasilffi/src/lib.rs
nunogois: You're talking about:
 - `count_toggle`
 - `count_variant`
 
Correct? What should happen in case we have an error there (e.g. one of the pointers is null)? From an interface perspective, I think it makes sense that we still let the caller know, even if it decides to just log it.

I'll merge this for now but I'm interested to know what you have in mind.

PR #48, opened 2023-10-10
fix: non strategy variants will work with empty strategy variants

sighphyre:
This allows base variants to resolve correctly even when strategy variants are defined as an empty list. Previously, if Yggdrasil found a strategy variant set that was an empty list, it would short circuit to the disabled variant and not fallback to checking the base variants


chriswk (approved):
Like the ratio of real change to added tests :)
Let's go

PR #49, opened 2023-10-12
feat: add dependent flags

thomasheartman:
This PR adds dependent flags to Yggdrasil, complying with the updated client specs.

Regarding these extra points not covered by the spec:

- metrics are not called on dependent features :: this is covered and tested
- impression events are called on dependent features for easier debugging :: yggdrasil doesn't touch impression events 
- warning events for missing dependencies are reported once :: yggdrasil doesn't deal with warnings

@kwasniew , there is a [line in `resolveVariant` in the Node SDK](https://github.com/Unleash/unleash-client-node/blob/6bed4e7f9f1993fc2b6232da8fbbda03cdb3973a/src/client.ts#L238) that checks for parent dependencies being satisfied. What exactly does it do? I tried doing something similar here, but removed it because the tests pass either way. Is this an edge case not covered by the tests, or is it just that the implementations differ here?

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -338,10 +338,10 @@ impl EngineState {
                             Some(variants) => {
                                 variants.contains(&self.get_variant(&parent.feature, &context).name)
                             }
-                            None => self.is_enabled(&parent.feature, &context),
+                            None => self.enabled(&parent_toggle, &context),
                         }
                     } else {
-                        !self.is_enabled(&parent.feature, &context)
+                        !self.enabled(&parent_toggle, &context)
                     }
                 })
             }
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -27,7 +27,7 @@ use unleash_types::client_metrics::{MetricBucket, ToggleStats};
 
 pub type CompiledState = HashMap<String, CompiledToggle>;
 
-pub const SUPPORTED_SPEC_VERSION: &str = "4.5.1";
+pub const SUPPORTED_SPEC_VERSION: &str = "4.5.2";
 
 pub struct CompiledToggle {
     pub name: String,
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -307,30 +307,34 @@ impl EngineState {
     }
 
     fn is_parent_dependency_satisfied(&self, toggle: &CompiledToggle, context: &Context) -> bool {
-        toggle.dependencies.iter().all(|parent| {
-            let Some(parent_toggle) = self.get_toggle(&parent.feature) else {
+        toggle.dependencies.iter().all(|parent_dependency| {
+            let Some(compiled_parent) = self.get_toggle(&parent_dependency.feature) else {
                 return false;
             };
 
-            if !parent_toggle.dependencies.is_empty() {
+            if !compiled_parent.dependencies.is_empty() {
                 return false;
             }
 
-            let parent_enabled = self.enabled(parent_toggle, context);
-            if parent.enabled.unwrap_or(true) {
-                match parent.variants.as_ref() {
-                    Some(variants) if !variants.is_empty() => {
-                        parent_enabled
-                            && self
-                                .check_variant_by_toggle(parent_toggle, context)
-                                .map(|variant| variants.contains(&variant.name))
-                                .unwrap_or(true)
-                    }
-                    _ => parent_enabled,
+            let parent_enabled = self.enabled(compiled_parent, context);
+            let expected_parent_enabled_state = parent_dependency.enabled.unwrap_or(true);
+            let parent_variant = self.check_variant_by_toggle(compiled_parent, context);
+
+            let is_variant_dependency_satisfied = {
+                if let (Some(expected_variants), Some(actual_variant)) =
+                    (&parent_dependency.variants, parent_variant)
+                {
+                    expected_variants.is_empty() || expected_variants.contains(&actual_variant.name)
+                } else {
+                    true
                 }
-            } else {
-                !parent_enabled
+            };
+
+            if !is_variant_dependency_satisfied {
+                return false;
             }
+
+            parent_enabled == expected_parent_enabled_state
         })
     }
 
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -29,23 +29,6 @@ pub type CompiledState = HashMap<String, CompiledToggle>;
 
 pub const SUPPORTED_SPEC_VERSION: &str = "4.5.1";
 
-#[derive(Debug, Serialize, Deserialize)]
-pub struct CompiledFeatureDependency {
-    pub feature: String,
-    pub enabled: Option<bool>,
-    pub variants: Option<Vec<String>>,
-}
-
-impl From<&FeatureDependency> for CompiledFeatureDependency {
-    fn from(value: &FeatureDependency) -> Self {
-        Self {
-            feature: value.feature.clone(),
-            enabled: value.enabled,
-            variants: value.variants.clone(),
-        }
-    }
-}
-
 pub struct CompiledToggle {
     pub name: String,
     pub enabled: bool,
@@ -54,7 +37,7 @@ pub struct CompiledToggle {
     pub variants: Vec<CompiledVariant>,
     pub impression_data: bool,
     pub project: String,
-    pub dependencies: Vec<CompiledFeatureDependency>,
+    pub dependencies: Vec<FeatureDependency>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -162,16 +145,10 @@ pub fn compile_state(state: &ClientFeatures) -> HashMap<String, CompiledToggle> 
     compiled_state
 }
 
-fn compile_dependencies(
-    dependencies: &Option<Vec<FeatureDependency>>,
-) -> Vec<CompiledFeatureDependency> {
-    if let Some(variants) = dependencies {
-        variants
-            .iter()
-            .map(CompiledFeatureDependency::from)
-            .collect()
-    } else {
-        vec![]
+fn compile_dependencies(dependencies: &Option<Vec<FeatureDependency>>) -> Vec<FeatureDependency> {
+    match dependencies {
+        Some(variants) => variants.clone(),
+        _ => vec![],
     }
 }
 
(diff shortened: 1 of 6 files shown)


unleash-yggdrasil/src/lib.rs
```diff
@@ -37,6 +37,7 @@ pub struct CompiledToggle {
     pub variants: Vec<CompiledVariant>,
     pub impression_data: bool,
     pub project: String,
+    pub dependencies: Vec<FeatureDependency>,
```
thomasheartman, line 40: I decided to follow the pattern of variants, where it's always a vec and we convert a `None` from the API into an empty Vec. Happy to go ahead and wrap it in an option if we think that's better.

thomasheartman:
All the clippy lints have been fixed, by the way.

kwasniew:
@thomasheartman this line in there because the variant code path probably didn't touch isDependenciesSatisfied earlier in the invocation call chain. If yggdrasil is calling isDependenciesSatisfied when doing getVariant then we're fine.

kwasniew (approved):
The algorithm looks correct. Let's see what @sighphyre says about Rust idioms since I can't comment on those.

thomasheartman:
> this line is there because the variant code path probably didn't touch isDependenciesSatisfied earlier in the invocation call chain. If yggdrasil is calling isDependenciesSatisfied when doing getVariant then we're fine.

Ah, got it! Yeah, get_variant uses the `enabled` method, which checks for dependencies, so that makes sense.

thomasheartman:
Ooh, I found a bug: the `get_variant` call counts the parent toggle if the child toggle specifies variant dependencies. I'll add a test and fix.

unleash-yggdrasil/src/lib.rs
```diff
@@ -303,9 +313,39 @@ impl EngineState {
...
+                !self.enabled(parent_toggle, context)
+            }
+        })
+    }
+
     fn enabled(&self, toggle: &CompiledToggle, context: &Context) -> bool {
-        let context = EnrichedContext::from(context.clone(), toggle.name.clone());
-        toggle.enabled && (toggle.compiled_strategy)(&context)
+        let enriched_context = EnrichedContext::from(context.clone(), toggle.name.clone());
+        match self.is_parent_dependency_satisfied(toggle, context) {
```
sighphyre, line 345: Matching booleans is weird. This can just be:

```
        toggle.enabled
            && self.is_parent_dependency_satisfied(toggle, context)
            && (toggle.compiled_strategy)(&enriched_context)
```

unleash-yggdrasil/src/lib.rs
```diff
@@ -1323,4 +1364,48 @@ mod test {
         assert_eq!(targeted_toggle.enabled, true);
         assert_eq!(targeted_toggle.variant.name, "another");
     }
+
+    #[test]
+    pub fn metrics_are_not_recorded_for_parent_flags() {
```
sighphyre, line 1369: Nice test!

unleash-yggdrasil/src/lib.rs
```diff
@@ -135,13 +137,21 @@ pub fn compile_state(state: &ClientFeatures) -> HashMap<String, CompiledToggle>
                 compiled_strategy: compile_rule(rule.as_str()).unwrap(),
                 impression_data: toggle.impression_data.unwrap_or_default(),
                 project: toggle.project.clone().unwrap_or("default".to_string()),
+                dependencies: compile_dependencies(&toggle.dependencies),
```
sighphyre, line 140: This is probably overkill, this can just be

```
dependencies: toggle.dependencies.clone().unwrap_or_default(),
```

Then we also don't need the `compile_dependencies` function

sighphyre, line 140: This is probably overkill, this can just be

Then we also don't need the `compile_dependencies` function

```suggestion
                dependencies: toggle.dependencies.clone().unwrap_or_default(),
```

unleash-yggdrasil/src/lib.rs
```diff
@@ -303,9 +313,39 @@ impl EngineState {
...
+            };
+
+            if !parent_toggle.dependencies.is_empty() {
+                return false;
+            }
+
+            if parent.enabled.unwrap_or(true) {
+                match parent.variants.as_ref() {
+                    Some(variants) if !variants.is_empty() => {
+                        variants.contains(&self.get_variant(&parent.feature, context).name)
```
sighphyre, line 333: This looks a little off to me. If we have variants we never check the parent's actual state, or did I miss something?

unleash-yggdrasil/src/lib.rs
thomasheartman: Sure thing!

unleash-yggdrasil/src/lib.rs
thomasheartman: Yup; good call.

unleash-yggdrasil/src/lib.rs
thomasheartman: Good question! I'll ... check. But that sounds like something the spec tests should have caught. I'll see what I can find.

unleash-yggdrasil/src/lib.rs
thomasheartman: handled in 5d58cb6857178c5669a3b6f948243c4c9120196d

unleash-yggdrasil/src/lib.rs
sighphyre: Yes, completely agree, the spec tests should have caught this but looking it over I don't think we have a case where we check both the constraint and a variant

thomasheartman: Wait, hang on; I think that checks out. Because we check whether parent flag's returned variant is one of the variants listed in the dependencies. If the parent flag is disabled, then the returned variant is the disabled variant, right? So implicitly, this assumes that it is.

Now, that *does* maybe open up the door for one edge case:
If the child depends on the parent being enabled and yielding the "disabled" variant. 
In that case, the dependency will be satisfied even if the parent flag isn't enabled 🤔

thomasheartman: But yeah, you're right! Added that as a test case to the spec in https://github.com/Unleash/client-specification/pull/65 (and locally) and I've fixed that point. Thanks 😄

thomasheartman:
In this last update, I have found a point which I think is correct, but I'd like to double check with @kwasniew before I merge:
To my understanding (from looking at the spec): If a child specifies a variant for its parent (e.g. `dependencies: [{ feature: 'parent', variants: ['whatever']}]`), and the parent is enabled, but has no variants, the child should *still* be enabled, right? Even if its variants constraints weren't satisfied?

This seems counterintuitive to me, but appears to align with what the spec says. Just wanna make sure this is good.

unleash-yggdrasil/src/lib.rs
```diff
@@ -303,9 +306,39 @@ impl EngineState {
...
+            let Some(parent_toggle) = self.get_toggle(&parent.feature) else {
+                return false;
+            };
+
+            if !parent_toggle.dependencies.is_empty() {
+                return false;
+            }
+
+            let parent_enabled = self.enabled(parent_toggle, context);
+            if parent.enabled.unwrap_or(true) {
```
sighphyre, line 320: I'm so sorry to do this, because I hate it when people trickle in PR comments over a few days, but yesterday was busy as hell and I didn't get time to give this section the love I wanted.

I feel this block is a little tough to read. The Rust is fine, but it captures a ton of domain logic cases in some very subtle ways through fallback values (like the unwraps or the catch all match arms).  I think we can make this a lot clearer and very obvious by breaking out the variant handling logic into it's own function. I'd like to suggest something like this:

```
    fn is_parent_dependency_satisfied(&self, toggle: &CompiledToggle, context: &Context) -> bool {
        toggle.dependencies.iter().all(|parent_dependency| {
            let Some(parent_toggle) = self.get_toggle(&parent_dependency.feature) else {
                return false;
            };

            if !parent_toggle.dependencies.is_empty() {
                return false;
            }

            let parent_enabled = self.enabled(parent_toggle, context);
            let parent_expectation = parent_dependency.enabled.unwrap_or(true);
            let parent_variant = self.check_variant_by_toggle(parent_toggle, context);

            if !self.is_variant_dependency_satisfied(&parent_dependency.variants, parent_variant) {
                return false;
            }

            parent_enabled == parent_expectation
        })
    }

    fn is_variant_dependency_satisfied(
        &self,
        expected_variants: &Option<Vec<String>>,
        actual_variant: Option<VariantDef>,
    ) -> bool {
        if let (Some(expected_variants), Some(actual_variant)) = (expected_variants, actual_variant)
        {
            expected_variants.is_empty() || expected_variants.contains(&actual_variant.name)
        } else {
            true
        }
    }
```

unleash-yggdrasil/src/lib.rs
sighphyre: I think this is okay. An Option would have made the actual implementation a little easier but Unleash is so fond of sending empty arrays for nulls that I think this is correct

sighphyre (commented):
I see the spec update. Nice one! I have one last input, outside of that I'm happy and willing to put an LGTM on it

unleash-yggdrasil/src/lib.rs
thomasheartman: No, I like that! How about using a local function, though? Doesn't feel like that's something that needs to leak out. But good point!

thomasheartman: I've attempted to address this in 78e6b36967eeb5a3f8833b27bbe8d7cc6d41e3e4, but I made `is_variant_dependency` a local variable instead. Let me know what you think.

As for why I made it a variable: this is state that is only relevant to this one method. By placing it in a separate method, it opens up for being called from other places. That's not necessarily bad, but it can be confusing when you're reading through this file and see this file. 

Next, I don't personally think that putting it in a separate function like that makes it much clearer than extracting it into a variable -- it *might* do the opposite (but I'm not sure, to be honest). 

However, if you disagree, I'm very happy to reconsider and move it into a function instead.

thomasheartman: As a follow-up, we *could* move the last two checks into one , but I'm not sure whether that helps or hinders clarity:

```rs
            is_variant_dependency_satisfied && parent_enabled == expected_parent_enabled_state
```

Thoughts?

sighphyre: I don't have strong feelings, it's private and well encapsulated enough and I wouldn't have thought twice about leaving it like this. Open to hiding it if you want though. There's a few things that can work here, in no particular order:

1. The method is small enough that it can be refactored into part of the parent method, the logical refactor around the way the domain logic is expressed is what I care about most
2. Nuno has a neat little pattern in the FFI layers where he uses an immediately executed closure for this: https://github.com/Unleash/yggdrasil/blob/26f86b6023fa8be52a20158aade78646550c00cf/yggdrasilffi/src/lib.rs#L133 that's an existing pattern so I'd lean on that over the local function. That may want an FnOnce though, so you'll need to try it if you want to go that route
3. Module level, pure function, which is what I usually lean on when I want to move something out of a struct impl
4. Local function is fine too, but I'd lean on (2) if you want that

I'm fine with any of the above or none, so I'll leave the choice in your hands

thomasheartman: Thanks! But just for clarity: I *did* make the changes. This comment seems to be addressing your suggestion? Or the way it was? I ended up just turning it into a named variable to make it clearer what it does and then using that. Are we talking past each other here?

sighphyre: Yeah, Github hadn't updated the conversation here so I was responding to a stale message.

I like what you did, gonna slap an LGTM on it

sighphyre (approved):
Nice! LGTM

PR #51, opened 2023-10-13
fix: allows current time to be assumed when calculating the context

sighphyre:
Fixes an issue where Yggdrasil won't assume the current time for a context field, instead it expects the SDKs to send it as part of the context, which they never do.

This patches this to allow Yggdrasil to assume a current time of now if the current time field is not overridden on the context field

### DIFF
--- a/unleash-yggdrasil/src/state.rs
+++ b/unleash-yggdrasil/src/state.rs
@@ -19,7 +19,10 @@ impl EnrichedContext {
             session_id: context.session_id.clone(),
             environment: context.environment.clone(),
             app_name: context.app_name.clone(),
-            current_time: context.current_time.clone(),
+            current_time: context
+                .current_time
+                .clone()
+                .or_else(|| Some(chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ").to_string())),
             remote_address: context.remote_address.clone(),
             properties: context.properties,
             toggle_name,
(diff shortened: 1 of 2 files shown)


unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -223,7 +223,7 @@ fn upgrade_constraint(constraint: &Constraint) -> String {
                     .collect::<Vec<String>>()
                     .join(", ")
             })
-            .unwrap_or_else(|| "".to_string());
+            .unwrap_or_default();
```
sighphyre, line 226: Not relevant here but Clippy is complaining and we like a happy Clippy

unleash-yggdrasil/src/state.rs
```diff
@@ -32,3 +35,49 @@ pub enum SdkError {
...
+mod test {
+    use crate::strategy_parsing::compile_rule;
+
+    use super::*;
+    use unleash_types::client_features::Context;
+
+    #[test]
+    fn converting_a_context_to_enriched_context_assumes_now_for_time_if_not_set() {
+        let context = Context::default();
+        let enriched_context = EnrichedContext::from(context, "test".into());
```
sighphyre, line 49: `EnrichedContext::from` is what Yggdrasil uses to convert a user facing context into its own internal extended representation

thomasheartman (approved):
I like it! I was gonna suggest a test to ensure the format of the date is correct, but that's covered by the test for using it in a constraint, so I think this is good 😄

PR #52, opened 2023-10-13
meta: add GitHub-cli as a dependency in devenv.nix

thomasheartman:
This adds the GitHub CLI as a dependency for nix users, so that we have the tools necessary for the release process.

Ideally, I'd like to add cargo smart-release, too, but it's not available on nixpkgs at the moment.


sighphyre (approved):
LGTM

PR #53, opened 2023-10-20
feat: improve cross platform interface, implement new yggdrasil metho…

daveleek:
improve cross platform interface, implement new yggdrasil methods and improve pointer handling

## About the changes

- Moves types into separate file
- Introduces interface and implementation per platform to add a more cross platform way of dealing with Yggdrasil libraries/builds
- Implements the new Count methods
- Implements the new get metrics method
- Improves memory/pointer handling
- Adds a simple memory leak probe for the .NET unleash engine

### DIFF
--- a/dotnet-engine/UnleashEngine.Tests/UnleashEngineTest.cs
+++ b/dotnet-engine/UnleashEngine.Tests/UnleashEngineTest.cs
@@ -44,7 +44,7 @@ public class Tests
 
             // Assert
             var diff = memoryTotal - baseline;
-            Assert.LessOrEqual(diff, 524288, process + " has a likely memory leak. Diff: " + diff + " bytes");
+            Assert.LessOrEqual(diff, 200000, process + " has a potential memory leak. Diff: " + diff + " bytes");
         };
 
         runTestFor(() => unleashEngine.IsEnabled("Feature.A", new Context()), "IsEnabled");
--- a/dotnet-engine/UnleashEngine.Tests/UnleashEngineTest.cs
+++ b/dotnet-engine/UnleashEngine.Tests/UnleashEngineTest.cs
@@ -5,6 +5,7 @@ using NUnit.Framework;
 using Unleash;
 using System;
 using Newtonsoft.Json.Linq;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 
 
 public class Tests
@@ -77,7 +78,7 @@ public class Tests
                 var toggleName = (string)test["toggleName"];
                 var expectedResult = (bool)test["expectedResult"];
 
-                var result = unleashEngine.IsEnabled(toggleName, context);
+                var result = unleashEngine.IsEnabled(toggleName, context) ?? false;
 
                 Assert.AreEqual(expectedResult, result, message: $"Failed client specification '{suite}': Failed test '{test["description"]}': expected {expectedResult}, got {result}");
             }
@@ -91,7 +92,7 @@ public class Tests
                 // Silly hack to apply formatting to the string from the spec
                 var expectedResult = JsonSerializer.Serialize(JsonSerializer.Deserialize<Variant>(test["expectedResult"].ToString(), options), options);
 
-                var result = unleashEngine.GetVariant(toggleName, context);
+                var result = unleashEngine.GetVariant(toggleName, context) ?? new Variant { Name = "disabled", Payload = null, Enabled = false };
                 var jsonResult = JsonSerializer.Serialize(result, options);
 
                 Assert.AreEqual(expectedResult, jsonResult, message: $"Failed client specification '{suite}': Failed test '{test["description"]}': expected {expectedResult}, got {result}");
--- a/dotnet-engine/UnleashEngine.Tests/UnleashEngineTest.cs
+++ b/dotnet-engine/UnleashEngine.Tests/UnleashEngineTest.cs
@@ -14,6 +14,45 @@ public class Tests
         PropertyNamingPolicy = JsonNamingPolicy.CamelCase
     };
 
+    [Test] 
+    public void MassTestMemoryUsage() {
+        // Arrange
+        var basePath = Path.Combine("..", "..", "..", "..", "..", "..", "client-specification", "specifications");
+        var suitePath = Path.Combine(basePath, "01-simple-examples.json");
+        var suiteData = JObject.Parse(File.ReadAllText(suitePath));
+
+        var unleashEngine = new UnleashEngine();
+        //unleashEngine.TakeState(suiteData["state"].ToString());
+
+        var runTestFor = (Action lambda, string process) => {
+
+            // Baseline / warm up
+            for (var i = 0; i < 1000000; i++) {
+                lambda();
+            }
+            GC.Collect();
+
+            var baseline = GC.GetTotalMemory(true);
+
+            // Act
+            for (var i = 0; i < 1000000; i++) {
+                lambda();
+            }
+            GC.Collect();
+
+            var memoryTotal = GC.GetTotalMemory(true);
+
+            // Assert
+            var diff = memoryTotal - baseline;
+            Assert.LessOrEqual(diff, 524288, process + " has a likely memory leak. Diff: " + diff + " bytes");
+        };
+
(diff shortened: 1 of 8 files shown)


dotnet-engine/UnleashEngine.Tests/UnleashEngineTest.cs
```diff
@@ -14,11 +14,50 @@ public class Tests
...
+            for (var i = 0; i < 1000000; i++) {
+                lambda();
+            }
+            GC.Collect();
+
+            var memoryTotal = GC.GetTotalMemory(true);
+
+            // Assert
+            var diff = memoryTotal - baseline;
+            Assert.LessOrEqual(diff, 524288, process + " has a likely memory leak. Diff: " + diff + " bytes");
```
sighphyre, line 47: Is this in bytes? So this is roughly half a meg?

dotnet-engine/UnleashEngine/Types.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    [JsonPropertyName("status_code")]
+    public string? StatusCode { get; set; }
+}
+
+public class EngineResponse<TResult> : EngineResponse {
+    public TResult? Value { get; set; }
+}
+
+
```
sighphyre, line 28:
```suggestion
```

dotnet-engine/UnleashEngine/Types.cs
```diff
@@ -0,0 +1,40 @@
...
+public class EngineResponse {
+    [JsonPropertyName("error_message")]
+    public string? ErrorMessage { get; set; }
+
+    [JsonPropertyName("status_code")]
+    public string? StatusCode { get; set; }
+}
+
+public class EngineResponse<TResult> : EngineResponse {
```
sighphyre, line 24: My .NET knowledge is failing me here, I don't actually know what this class declaration does O.O

dotnet-engine/UnleashEngine.Tests/UnleashEngineTest.cs
daveleek: Yeah my brain's trapped in the pre MB/MiB times - so this is half a MiB. I'll change to half a MB.
But yes. It's heavily influenced by the Ruby implementation - the real diff appears to be around a tenth of that (or a hundredth when the engine actually has state first)

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -5,85 +5,95 @@
...
+            return;
+        }
+
+        var takeStateJson = Marshal.PtrToStringUTF8(takeStatePtr);
+
+        var takeStateResult = takeStateJson != null ?
+            JsonSerializer.Deserialize<EngineResponse>(takeStateJson, options) :
+            null;
+
+        platformEngine.FreeResponse(takeStatePtr);
```
sighphyre, line 58: These free responses need to be as close to the reading of the pointer as possible 

(looks to be this line: `var takeStateJson = Marshal.PtrToStringUTF8(takeStatePtr);`)

Line 54 can potentially fail, which will cause a leak, as written, can we move this line to line 53? If we can't, we probably gotta wrap this in a try/finally

Same thing for other methods that release the pointers

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -5,85 +5,95 @@
...
+
+        if (takeStateResult?.StatusCode == "Error") {
+            throw new Exception($"Error: {takeStateResult?.ErrorMessage}");
```
sighphyre, line 61: Can we make an Exception type for this? I hate having to catch `Exception`

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -5,85 +5,95 @@
...
+        }
+        else
+        {
+            var isEnabledJson = Marshal.PtrToStringUTF8(isEnabledPtr);
+
+            var isEnabledResult = isEnabledJson != null ?
+                JsonSerializer.Deserialize<EngineResponse<bool?>>(isEnabledJson, options) :
+                null;
+
+            platformEngine.FreeResponse(isEnabledPtr);
```
sighphyre, line 82: Same as earlier comment

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -5,85 +5,95 @@
...
     {
         string contextJson = JsonSerializer.Serialize(context, options);
-        return engine_is_enabled(state, toggleName, contextJson);
+        var isEnabledPtr = platformEngine.CheckEnabled(state, toggleName, contextJson);
+
+        if (isEnabledPtr == IntPtr.Zero)
+        {
+            return false;
+        }
+        else
```
sighphyre, line 74: The `else` seems to cover all the other logic, can we just remove the `else`?

dotnet-engine/UnleashEngine/IFFIAccess.cs
```diff
@@ -0,0 +1,21 @@
+namespace Unleash;
+
+internal interface IFFIAccess {
```
sighphyre, line 3: Can we try golf this together next week? We should try a bit to get this better

Is it a blocker to leave this open and work against a dev version on the SDK side?

dotnet-engine/UnleashEngine/UnleashEngine.cs
daveleek: Thank you, valuable input! I'm fairly certain it can be moved!

sighphyre (changes requested):
Overall looks great, I think you've gotten a pretty nice wrapper here.

I have a few critical memory things around the way pointers are released here and some small cleanups.

I'd like to take a stab together in a pairing session to see what we can do about the OS specific FFI interfaces

dotnet-engine/UnleashEngine/IFFIAccess.cs
daveleek: Absolutely! The intention behind this is to have something that works well but at the same time gets the conversation going

daveleek:
Thank you for a good review @sighphyre! A pairing session next week sounds great, and I'll address the other stuff in the mean time

dotnet-engine/UnleashEngine.Tests/UnleashEngineTest.cs
sighphyre: Let's be more aggressive then! If .NET is better than Ruby with memory (really, really not surprising), let's get this closer to the real number!

dotnet-engine/UnleashEngine/UnleashEngine.cs
```diff
@@ -5,99 +5,142 @@
...
-            return variant;
+        var variantResult = variantJson != null ?
+            JsonSerializer.Deserialize<EngineResponse<Variant>>(variantJson, options) :
+            null;
+
+        if (variantResult?.StatusCode == "Error") {
+            throw new UnleashException($"Error: {variantResult?.ErrorMessage}");
         }
+
+        var variant = variantResult?.Value ?? Variant.DISABLED_VARIANT;
```
gastonfournier, line 117: Here we should not return the disabled variant because the user can provide a default value when the feature is not found. So it should be an SDK responsibility. We should either return null or an error. Most likely a null

sighphyre: Ooo good spot!

gastonfournier: Actually, I'm thinking about this again, and if UnleashEngine does not behave as an SDK and we're saying that the SDK should take care of some details, then how can we test it against the specification?

daveleek: > then how can we test it against the specification

Yeah this is what caused me to return a default disabled variant, spec was failing.

I wonder if we on the SDK-side should check if it's the disabled variant being returned and if so and a default has been requested return the default. But I'm open to suggestions

sighphyre (approved):
Still stuff to do here but I don't want this to get stale and make life hard. Wanna merge this?

PR #55, opened 2023-10-25
chore: make Java Engine pass specs

gastonfournier:
## About the changes
Adapt Java Engine to latest changes in Yggdrasil and make tests pass the specification

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
+++ b/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
@@ -3,13 +3,14 @@
  */
 package io.getunleash.engine;
 
-import java.nio.file.Paths;
-
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.JsonMappingException;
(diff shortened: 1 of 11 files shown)


java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -25,40 +26,62 @@ interface UnleashFFI extends Library {
 
     Pointer take_state(Pointer ptr, String toggles);
 
-    boolean check_enabled(Pointer ptr, String name, String context);
+    Pointer check_enabled(Pointer ptr, String name, String context);
 
     Pointer check_variant(Pointer ptr, String name, String context);
```
sighphyre, line 32: We'll need the metrics methods here too but that can be in another PR

sighphyre (approved):
Yeah this is really nice, the error handling is great

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
gastonfournier: Yes, will cover that here: https://linear.app/unleash/issue/2-1560/implement-metrics

PR #56, opened 2023-10-26
feat: custom strategies

sighphyre:
This is a barebones implementation of custom strategies in the engine code and an implementation in Ruby. 

# Why

Custom strategies are a bit of a devil but they're a devil that's here to stay for at least a year or two. Because of this, the implementation leans towards something that can easily be removed later, rather than something that's polished. Most of the hacky stuff lives in the host language code and not in Ygg itself This gets us support for the next year or two without causing damage to the grammar.

# How

## Engine code

The engine code works by exposing a grammar fragment called "external_value["some-lookup-name"]. This allows us to inject a mapping of name -> boolean pairs into the grammar and flip bits of the logic on or off at runtime. So a rule that looks like:

`user_id > 7 and external_value["some-lookup"]`

Will resolve to 

`user_id > 7 and true`

If we pass the following:

{
  "some-lookup": true
}

I believe this is a generally useful thing and is not a hack but I'm open to being told I'm wrong here.

## Host language code

This is definitely a hack. We allow the user to `register_custom_strategies(some_list)`, which allows them to pass an arbitrary class instance into the host language code. When receiving a blob of feature toggle state, we first find all the custom strategies for each toggle. When we execute is_enabled, we check if we have any custom strategies for that toggle name. If we do, we evaluate them in the *host* code and pass the results to the engine, allowing "external_value" to deal with it. This allows us to respect constraints on the custom strategy, which isn't really possible any other way (or at least not one I can think of that's not significantly worse). 

# The Ugly

- To get and track the toggles with custom strategies, we need to parse the JSON blob in the host language. But we also need to do that in Ygg itself. So we incur the cost of parsing twice
- This depends on a convention between engine and host, where the host passes "customStrategy1", customStrategy2", etc to the engine and the engine reads them back in that order. It works but it is a convention that needs to be respected
- Custom strategies are ignored if not present. That's what our docs say we should do but this opens doors to debugging hell 
- Custom strategies are a no-op for parent feature toggles. It's too ugly to keep track of it all. Quick conversation with the devs on that feature suggests it's okay to do this and block this in the UI


node-sdk/src/lib.rs
```diff
@@ -30,13 +30,13 @@ impl UnleashEngine {
     #[wasm_bindgen(method, js_name = isEnabled)]
     pub fn is_enabled(&self, name: String, context: &JsValue) -> bool {
         let context: InnerContext = serde_wasm_bindgen::from_value(context.clone()).unwrap();
-        self.engine_state.is_enabled(&name, &context)
+        self.engine_state.is_enabled(&name, &context, None)
```
sighphyre, line 33: The WASM compiles as part of the main build. This is just to not block the build

ruby-engine/scripts/mem_check.rb
```diff
@@ -6,34 +6,34 @@
...
-    diff = post_check_memory - pre_check_memory
-    if diff < 0.5
-        puts "#{method_name_being_tested} Memory diff is within half a MB, this is an expected range"
-    else
-        STDERR.puts "WARNING: LIKELY MEMORY LEAK DETECTED. THIS REQUIRES HUMAN ATTENTION"
-        STDERR.puts "#{method_name_being_tested} Memory diff is not within the expected half MB range, this likely indicates a leak within the engine"
-    end
-
-    puts "---"
+  puts "#{method_name_being_tested} Warming up"
```
sighphyre, line 9: No changes n this file. Just a second sweep with the formatter apparently

ruby-engine/unleash-engine.gemspec
```diff
@@ -1,16 +1,16 @@
 Gem::Specification.new do |s|
-    s.name        = 'unleash-engine'
```
sighphyre, line 2: Also formatter, sorry

test-data/simple.json
```diff
@@ -25,6 +25,54 @@
       "name": "Feature.C",
       "enabled": true,
       "strategies": []
+    },
```
sighphyre, line 28: We don't have any client spec tests for custom strats, it's convenient to have these for tests

ruby-engine/spec/custom_strategy_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+      expect(strategy_results["customStrategy1"]).to eq(true)
+      expect(strategy_results["customStrategy2"]).to eq(true)
+    end
+
+    it 'returns false for missing implementations' do
+      handler.register_custom_strategies([])
+      strategy_results = handler.evaluate_custom_strategies("Feature.A", {})
+      expect(strategy_results.length).to eq(2)
+      expect(strategy_results["customStrategy1"]).to eq(false)
+      expect(strategy_results["customStrategy1"]).to eq(false)
```
gastonfournier, line 86: Is it intentional to add twice the same check? Otherwise maybe you want:
```suggestion
      expect(strategy_results["customStrategy2"]).to eq(false)
```

unleash-yggdrasil/src/lib.rs
```diff
@@ -323,7 +323,7 @@ impl EngineState {
                 return false;
             }
 
-            let parent_enabled = self.enabled(compiled_parent, context);
+            let parent_enabled = self.enabled(compiled_parent, context, None); //parent toggles explicitly don't support custom strategies
```
gastonfournier, line 326: Is it a design decision? I'm thinking on what should we do in other Engines

unleash-yggdrasil/src/lib.rs
```diff
@@ -345,25 +345,35 @@ impl EngineState {
...
+        &self,
+        context: &Context,
+        external_values: Option<HashMap<String, bool>>,
+    ) -> Option<HashMap<String, ResolvedToggle>> {
         self.compiled_state.as_ref().map(|state| {
             state
                 .iter()
                 .map(|(name, toggle)| {
-                    let enabled = self.enabled(toggle, context);
+                    let enabled = self.enabled(toggle, context, external_values.clone());
```
gastonfournier, line 370: Should we borrow instead of cloning? I'm thinking this should probably be immutable at this point

unleash-yggdrasil/src/lib.rs
```diff
@@ -372,26 +382,41 @@ impl EngineState {
...
+    pub fn resolve(
+        &self,
+        name: &str,
+        context: &Context,
+        external_values: Option<HashMap<String, bool>>,
+    ) -> Option<ResolvedToggle> {
         self.compiled_state.as_ref().and_then(|state| {
             state.get(name).map(|compiled_toggle| ResolvedToggle {
-                enabled: self.enabled(compiled_toggle, context),
+                enabled: self.enabled(compiled_toggle, context, external_values.clone()),
```
gastonfournier, line 393: Same here, not sure if we can avoid cloning here

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -372,6 +372,18 @@ fn list_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
...
+    let strategy_index = string(node.next().unwrap());
+    Box::new(move |context| {
+        context
+            .external_results
+            .as_ref()
+            .and_then(|strategy_results| strategy_results.get(&strategy_index))
+            .copied()
+            .unwrap_or(false)
+    })
+}
```
gastonfournier, line 385: I'm having a hard time at how we should handle inverted but if we're consciously dropping it, we should at least state why

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -827,4 +843,42 @@ mod tests {
         let rule = "user_id in [\"Nobody likes \\\"scare quotes\\\"\"]";
         compile_rule(rule).unwrap();
     }
+
+    #[test]
+    fn missing_external_value_produces_false_without_error() {
+        let rule = "external_value[\"i_do_not_exist\"]";
```
gastonfournier, line 849:
```
let rule = "!external_value[\"i_do_not_exist\"]";
```
would this be a good test for inverted?

unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -6,13 +6,43 @@ use crate::state::SdkError;
...
+enum StrategyType {
+    Default,
+    UserWithId,
+    GradualRolloutUserId,
+    GradualRolloutSessionId,
+    GradualRolloutRandom,
+    FlexibleRollout,
+    RemoteAddress,
+    Custom(String),
+}
```
gastonfournier, line 18: Is this accessible somehow from the Engine (or maybe the is_custom function that provided a strategy name, it returns true or false whether its custom or not)? I'm thinking of a centralized list of strategies. Maybe something for later

gastonfournier (approved):
Looks good to me, I left a couple of questions but I think they're potentially future improvements. I've been also thinking of potential ways of making it better but I can't think of any atm

ruby-engine/spec/custom_strategy_spec.rb
sighphyre: *cough* duplicated line, nice catch, I'll remove

unleash-yggdrasil/src/lib.rs
sighphyre: Borrow vs owned is more about memory management than mutability but this is good feedback, this doesn't need to be a clone at this point. I've made this propagate via a borrow all the way down to the where the enriched context is build, which does require an owned type

unleash-yggdrasil/src/lib.rs
sighphyre: We can and we should! Good spot, it's borrowed now!

unleash-yggdrasil/src/strategy_upgrade.rs
sighphyre: Yuh, thought about it too, it's good feedback. Currently not accessible but we can make it so in the future. Probably want to do that in another PR though, this one is too big already

unleash-yggdrasil/src/strategy_parsing.rs
sighphyre: I've added specific tests for this. One to check that inversion works on `external_value` and a second to prove that it remains unchanged when its not found (just to match the current state of Ygg)

PR #57, opened 2023-10-26
chore: use an instance wrapper of Yggdrasil

gastonfournier:
This small refactor extracts Yggdrasil FFI into an internal class enabling us to provide the path of the compiled version of Yggdrasil in the tests.

It adds a thin layer on top of the engine pointer which will enable us to free it up properly in a later PR


java-engine/src/main/java/io/getunleash/engine/YggdrasilFFI.java
```diff
@@ -0,0 +1,72 @@
...
+
+        this.ffi = Native.load(combinedPath, UnleashFFI.class);
+        this.enginePtr = this.ffi.new_engine();
+    }
+
+    Pointer takeState(String toggles) {
+        return this.ffi.take_state(this.enginePtr, toggles);
+    }
+
+    void free() {
```
sighphyre, line 57: This little beast has to get called when this falls out of scope, otherwise we end up with a memory leak. ChatGPT says a Cleaner class is the right way to go here but I'm not very familiar with that.

sighphyre (approved):
I like the refactor! LGTM and all that.

This code definitely leaks the engine pointers, but I think we should merge this and deal with that later

java-engine/src/main/java/io/getunleash/engine/YggdrasilFFI.java
gastonfournier: Yes, I wanted to solve that in a separate PR because that's already how it is

PR #58, opened 2023-10-26
chore: implement memory clean up

gastonfournier:
Autoclosable implementation and some tests.
Automated testing with GH Actions

References:
1. https://inside.java/2022/05/27/testing-clean-cleaner-cleanup/
2. https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/Cleaner.html

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/YggdrasilFFI.java
+++ b/java-engine/src/main/java/io/getunleash/engine/YggdrasilFFI.java
@@ -22,7 +22,7 @@ interface UnleashFFI extends Library {
     void free_response(Pointer pointer);
 }
 
-class YggdrasilFFI {
+class YggdrasilFFI implements AutoCloseable {
     private final UnleashFFI ffi; // instance ffi
     private final Pointer enginePtr;
 
@@ -54,10 +54,6 @@ class YggdrasilFFI {
         return this.ffi.take_state(this.enginePtr, toggles);
     }
 
-    void free() {
-        this.ffi.free_engine(this.enginePtr);
-    }
-
     void freeResponse(Pointer response) {
         this.ffi.free_response(response);
     }
@@ -68,5 +64,10 @@ class YggdrasilFFI {
 
     Pointer checkVariant(String name, String context){
         return this.ffi.check_variant(this.enginePtr, name, context);
+    }
+
+    @Override
+    public void close() {
+        this.ffi.free_engine(this.enginePtr);
     }
 }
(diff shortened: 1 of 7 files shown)


java-engine/src/main/java/io/getunleash/engine/YggdrasilFFI.java
```diff
@@ -45,19 +50,27 @@ class YggdrasilFFI {
...
         String combinedPath = Paths.get(libraryPath, libImpl).toString();
+        return Native.load(combinedPath, UnleashFFI.class);
+    }
 
-        this.ffi = Native.load(combinedPath, UnleashFFI.class);
+    YggdrasilFFI(String libraryPath) {
+        this(loadLibrary(libraryPath));
+    }
+
+    YggdrasilFFI(UnleashFFI ffi) {
```
gastonfournier, line 60: This improves testing as we can now inject an FFI mocked instance

java-engine/src/test/java/io/getunleash/engine/YggdrasilFFITest.java
```diff
@@ -0,0 +1,117 @@
...
+            new YggdrasilFFI(absoluteValidPath());
+        });
+
+        assertThrows(UnsatisfiedLinkError.class, () -> {
+            new YggdrasilFFI("/non/existent/path");
+        });
+    }
+
+    @Test
+    void testResourceCleanup() {
```
gastonfournier, line 54: This test is important, for details about how/why this works, take a deep breath and look into https://inside.java/2022/05/27/testing-clean-cleaner-cleanup/

java-engine/src/main/java/io/getunleash/engine/YggdrasilFFI.java
```diff
@@ -69,4 +82,20 @@ Pointer checkEnabled(String name, String context){
     Pointer checkVariant(String name, String context){
         return this.ffi.check_variant(this.enginePtr, name, context);
     }
+
+    private static final class YggdrasilNativeLibraryResourceCleaner implements Runnable {
```
sighphyre, line 86: A whole Runnable for this?

Java, you've been drinking again, haven't you?

Jokes aside, this looks sane

sighphyre (approved):
LGTM

java-engine/src/main/java/io/getunleash/engine/YggdrasilFFI.java
gastonfournier: Yeah, also because of:
> The cleaning action could be a lambda but all too easily will capture the object reference, by referring to fields of the object being cleaned, preventing the object from becoming phantom reachable. Using a static nested class, as above, will avoid accidentally retaining the object reference.

PR #59, opened 2023-10-30
chore: add metrics support

gastonfournier:
- Follow recommendations from Jackson for performance.
- Ability to count enabled/disabled features and variants.
- Ability to pull metrics.
- Expose some classes that will be accessed from the SDK and removed ones having business logic that we want on the SDK


java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -3,21 +3,28 @@
...
         this(new YggdrasilFFI());
     }
 
     UnleashEngine(YggdrasilFFI yggdrasil) {
         this.yggdrasil = yggdrasil;
-        mapper = new ObjectMapper();
+        ObjectMapper mapper = new ObjectMapper();
         mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        reader = mapper.reader();
+        writer = mapper.writer();
```
gastonfournier, line 27: Following https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -27,25 +36,36 @@ public void takeState(String toggles) throws YggdrasilInvalidInputException {
         }
     }
 
-    public boolean isEnabled(String name, Context context) throws YggdrasilInvalidInputException {
+    public IsEnabledResponse isEnabled(String name, Context context) throws YggdrasilInvalidInputException {
```
gastonfournier, line 39: Let the SDK handle this, otherwise, we won't be able to inject a default value.

I'm still wondering if this method should accept an additional parameter to encapsulate the logic of handling isValid and reduce the exposure of internal classes, but I think I will only know when I start to integrate this into Java SDK.

java-engine/src/main/java/io/getunleash/engine/VariantDef.java
```diff
@@ -3,7 +3,7 @@
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-class VariantDef {
+public class VariantDef {
```
gastonfournier, line 6: Should be accessible from the SDK code

java-engine/src/main/java/io/getunleash/engine/VariantResponse.java
```diff
@@ -3,7 +3,7 @@
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-class VariantResponse extends YggResponse<VariantDef> {
+public class VariantResponse extends YggResponse<VariantDef> {
```
gastonfournier, line 6: Should be accessible from the SDK code

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -1,62 +1,86 @@
...
+            String jsonContext = writer.writeValueAsString(context);
+            YggResponse<Boolean> isEnabled = read(yggdrasil.checkEnabled(name, jsonContext), new TypeReference<>() {});
+            return isEnabled.getValue();
         } catch (JsonProcessingException e) {
             throw new YggdrasilInvalidInputException(context);
         }
     }
 
-    public VariantResponse getVariant(String name, Context context) throws YggdrasilInvalidInputException {
+    public VariantDef getVariant(String name, Context context) throws YggdrasilInvalidInputException, YggdrasilError {
```
gastonfournier, line 50: I'm not sure if I want to start having too many exceptions in this interface... :thinking: 
- `YggdrasilInvalidInputException`: for bad user input
- `YggdrasilError`: something went terribly wrong

I believe I can just have YggdrasilError and have YggdrasilInvalidInputException extending from it, but I'd like to get your input @daveleek @sighphyre

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -1,62 +1,86 @@
...
-    private <T> T read(Pointer pointer, Class<T> clazz) {
+    private <T> T read(Pointer pointer, TypeReference<T> typeReference) {
         String str = pointer.getString(0, UTF_8);
         yggdrasil.freeResponse(pointer);
         try {
             System.out.println(str); // TODO use a logging library. SLF4J?
-            return mapper.readValue(str, clazz);
-        } catch (JsonProcessingException e) {
-            throw new YggdrasilParseException(str, clazz, e);
+            return reader.forType(typeReference).readValue(str);
```
gastonfournier, line 81: Potential for optimization:
- reader.forType(typeReference)
could be in a Map for each one of the known types...

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
daveleek: Maybe there's a tiny discussion to be had around whether or not we use the name "Yggdrasil" in the wrapper/platform specific engine - and leaking it out to SDKs. That it should be the same across all wrappers and SDKs?

I know TakeState will throw a bad json error on TakeState, would that be happening here as well with the Context @sighphyre ?

gastonfournier: WRT takeState and bad input I'm mapping this into a checked exception: YggdrasilInvalidInputException. Not sure we want standardization across Engines at this stage.

About Yggdrassil leaking outside, I like to make it visible, but also understand your reasons, not sure...

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
sighphyre: Oh! I didn't know this. I learned something today

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
sighphyre: Yeah take state can go bad in theory but it _shouldn't_ since we control both sides. If we're worried about too many exceptions for folks to have to think about we can always have one base Yggdrasil exception class and have child classes. Gives people enough granularity to catch specific things if they care and anything if they don't

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -1,62 +1,86 @@
...
+        } catch (IOException e) {
+            throw new YggdrasilParseException(str, typeReference.getClass(), e);
```
sighphyre, line 83: Doesn't this destroy the original stack trace? Can we include the caught exception here?

sighphyre (approved):
Yeah LGTM!

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
gastonfournier: Ok, I'll keep it this way for now, as it will only be consumed by our SDK which we control. If we feel the pain, we change it later. Maybe next iteration we can have the hierarchy of exceptions. I think that'd be a good solution

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
gastonfournier: It shouldn't as it passes it as the third parameter to YggdrasilParseException

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
chriswk: So did I, this feels like it has changed since I learned Jackson, I see the wiki page was updated in 2019, so definitely long after I learned Jackson. Cool. TIL :)

Issue #60, opened 2023-10-30
Evaluating strategy variants does not utilize the groupId for evaluation


sjaanus:
### Describe the bug

With the introduction of strategy variants, it's essential to use the groupId from the strategy for variant selection. Currently, only the feature name is being used, neglecting the groupId.

### Steps to reproduce the bug

Upgrade to 5.0.2 client spec and it will reproduce.

### Subscription type

None

### Hosting type

None

PR #61, opened 2023-10-30
feat: new hashing implemented for variants

sjaanus:
1. Variants will be hashed with new seed
2. Fixed a [bug](https://github.com/Unleash/yggdrasil/issues/60), where variant strategy evaluation was not using groupId from strategy

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -1167,15 +1167,15 @@ mod test {
 
     // The client spec doesn't actually enforce anything except userId for variant overrides, so this is
     // getting its own test set until the client spec can take over that responsibility
-    #[test_case("userId", "7", & ["7"], true; "Basic example")]
-    #[test_case("userId", "7", & ["7", "8"], true; "With multiple values")]
-    #[test_case("userId", "7", & ["2", "9"], false; "Expected not to match against missing property")]
-    #[test_case("sessionId", "7", & ["2", "7"], true; "Resolves against session id")]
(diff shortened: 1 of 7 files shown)


unleash-yggdrasil/src/lib.rs
```diff
@@ -27,13 +28,15 @@ use unleash_types::client_metrics::{MetricBucket, ToggleStats};
...
+pub const SUPPORTED_SPEC_VERSION: &str = "5.0.2";
+const VARIANT_NORMALIZATION_SEED: u32 = 86028157;
+
 
 pub struct CompiledToggle {
     pub name: String,
     pub enabled: bool,
     pub compiled_strategy: RuleFragment,
-    pub compiled_variant_strategy: Option<Vec<(RuleFragment, Vec<CompiledVariant>)>>,
+    pub compiled_variant_strategy: Option<Vec<(RuleFragment, Vec<CompiledVariant>, String)>>,
```
sjaanus, line 39: When we are compiling variant strategy, we also need to pull out the groupId from strategy, not only variants.

unleash-yggdrasil/src/lib.rs
```diff
@@ -394,8 +398,8 @@ impl EngineState {
 
     fn resolve_variant<'a>(
         &self,
-        toggle: &CompiledToggle,
```
sjaanus, line 397: Resolving does not need full toggle, it just needs groupId.

sighphyre (approved):
LGTM

gardleopard (approved):
LGTM

PR #65, opened 2023-11-01
refactor: fix normalized hash result to start from 1 instead of 0

sjaanus:
To align it with rest of SDKs, hash result should start from index 1 instead of 0.


gardleopard (approved):
LGTM

sighphyre (approved):
So not a logic change after all

Sure, I don't have strong feelings here

PR #66, opened 2023-11-07
feat: wasm engine for direct grammar access

sighphyre:
This adds a spike on a WASM engine for direct access to the grammar engine. This is primarily to support the Open Feature initiative to be able to spike on the grammar

This also removes the old node-sdk, which didn't work correctly in its current shape and swaps that out for a barebones implementation for only what we need


chriswk (approved):
Cool! LGTM

wasm-engine/README.md
```diff
@@ -0,0 +1,84 @@
...
+First, install the package:
+
+```sh
+$ yarn add @unleash/yggdrasil-engine
+```
+
+Then, you can use it in your code:
+
+```js
+  const yggdrasil = require("../pkg/wasm_engine.js");
```
gastonfournier, line 22: Is this correct? Or should it be 
```suggestion
  const yggdrasil = require("../pkg/wasm-engine.js");
```
I'm saying because that's the package name but I can't find where we make it snake case

gastonfournier (approved):
I like it and also looks good code-wise just a small potential typo

daveleek (approved):
LG, but you should also get someone more qualified to have a look :)

wasm-engine/README.md
sighphyre: Good catch but this is compiler magic that does the case translation. This is just a placeholder so I can do a build pipeline in the next pass, apologies, should have pointed that out

wasm-engine/Cargo.toml
```diff
@@ -11,25 +11,27 @@ crate-type = ["cdylib", "rlib"]
...
-serde-wasm-bindgen = "0.4.5"
-wasm-bindgen = "0.2.83"
+getrandom = { version = "0.2", features = ["js"] }
 
 # The `console_error_panic_hook` crate provides better debugging of panics by
 # logging them with `console.error`. This is great for development, but requires
 # all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
 # code size when deploying.
-console_error_panic_hook = {version = "0.1.6", optional = true}
+console_error_panic_hook = { version = "0.1.1", optional = true }
```
chriswk, line 22: Why the downgrade here?

wasm-engine/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+    #[serde(rename = "appName")]
+    pub app_name: Option<String>,
+    #[serde(rename = "currentTime")]
+    pub current_time: Option<String>,
+    #[serde(rename = "remoteAddress")]
+    pub remote_address: Option<String>,
+    #[serde(rename = "groupId")]
+    pub group_id: Option<String>,
+    #[serde(rename = "properties")]
+    pub properties: Option<HashMap<String, String>>,
```
chriswk, line 28: #[serde(rename_all = "camelCase")] and skip the explicit rename for all of them?

wasm-engine/Cargo.toml
sighphyre: Good spot, no good reason, this was modified by a tool, I've reverted it to the previous version

wasm-engine/src/lib.rs
sighphyre: Ooo thanks! No idea why I went the long way here, applied rename_all

PR #70, opened 2023-11-07
refactor: rework dotnet internals so we don't have to have a class for each arch

sighphyre:
This reworks the dotnet internals so that we don't have to have a class for each architecture type. This provides a bunch of null static delegates, which are filled in by a static constructor. This lets us have a single point of change (yay!)


dotnet-engine/Yggdrasil.Engine/FFI.cs
```diff
@@ -0,0 +1,152 @@
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace Yggdrasil;
+
+internal static class FFI
```
sighphyre, line 6: I've moved this to a static class, there's no need to hold state for this, that's done by the engine pointer

dotnet-engine/Yggdrasil.Engine/FFI.cs
```diff
@@ -0,0 +1,152 @@
...
+            libraryName = "libyggdrasilffi.so";
+        else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            libraryName = "libyggdrasilffi.dylib";
+        else
+            throw new PlatformNotSupportedException("Unsupported platform");
+
+        string assemblyLocation = Assembly.GetExecutingAssembly().Location;
+        string assemblyDirectory = Path.GetDirectoryName(assemblyLocation);
+
+        return Path.Combine(assemblyDirectory, libraryName);
```
sighphyre, line 95: To make the tests pick up the execution location so they can find the so/dll/dylib

daveleek (approved):
LGTM! I can't fault this

I had no idea you could load things this way, but I'm glad we can and I'm glad you found it! :)

PR #72, opened 2023-11-07
refactor: no warnings please

sighphyre:
Removes a bunch of places where the compiler was complaining about potential nulls and seals a few publicly facing classes. Also adds a hashcode method where we implemented equals

Just future proofing before we get runnaway compiler warnings and it becomes hard to fix

### DIFF
--- a/dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
+++ b/dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
@@ -77,7 +77,7 @@ public class Tests
                 var toggleName = (string)test["toggleName"];
                 var expectedResult = (bool)test["expectedResult"];
 
-                var result = yggdrasilEngine.IsEnabled(toggleName, context) ?? false;
+                var result = yggdrasilEngine.IsEnabled(toggleName, context!) ?? false;
 
                 Assert.AreEqual(expectedResult, result, message: $"Failed client specification '{suite}': Failed test '{test["description"]}': expected {expectedResult}, got {result}");
             }
@@ -91,7 +91,7 @@ public class Tests
                 // Silly hack to apply formatting to the string from the spec
                 var expectedResult = JsonSerializer.Serialize(JsonSerializer.Deserialize<Variant>(test["expectedResult"].ToString(), options), options);
 
-                var result = yggdrasilEngine.GetVariant(toggleName, context) ?? new Variant { Name = "disabled", Payload = null, Enabled = false };
+                var result = yggdrasilEngine.GetVariant(toggleName, context!) ?? new Variant { Name = "disabled", Payload = null, Enabled = false };
                 var jsonResult = JsonSerializer.Serialize(result, options);
 
                 Assert.AreEqual(expectedResult, jsonResult, message: $"Failed client specification '{suite}': Failed test '{test["description"]}': expected {expectedResult}, got {result}");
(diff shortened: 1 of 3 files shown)


dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
```diff
@@ -77,7 +77,7 @@ public void TestClientSpec()
                 var toggleName = (string)test["toggleName"];
                 var expectedResult = (bool)test["expectedResult"];
 
-                var result = yggdrasilEngine.IsEnabled(toggleName, context) ?? false;
+                var result = yggdrasilEngine.IsEnabled(toggleName, context!) ?? false;
```
sighphyre, line 80: Just a test, this can never be null or the test will explode, telling the compiler to hush here feels okay

dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Yggdrasil;
 
-public class Context
+public sealed class Context
```
sighphyre, line 5: I've marked a bunch of publicly exposed classes as sealed. This is solving 0 problems outside of crazy people deriving these classes but I'd rather stop them now than have to stop them later. If they never do, then this does nothing

dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -13,29 +13,31 @@ public class Context
...
+public class EngineResponse<TValue> : EngineResponse
+{
     public TValue? Value { get; set; }
 }
 
-public class Variant
+public sealed class Variant
 {
-    public string Name { get; set; }
+    public string Name { get; set; } = null!;
```
sighphyre, line 32: Sneaky .NET syntax - we assign a default value of null but tell the compiler that this will never be the case. I think this is okay here since these are always constructed through deserialisation

dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -45,25 +47,35 @@ public override bool Equals(object? obj)
         var payload = (Payload)obj;
         return Value == payload.Value && PayloadType == payload.PayloadType;
     }
+
+    public override int GetHashCode()
```
sighphyre, line 51: We have equals. We need hashcode as well to prevent the compiler complaining. I've picked an unchecked here, which allows the integer to wrap rather than creating an overflow exception

dotnet-engine/Yggdrasil.Engine/Types.cs
gastonfournier: weird, maybe @daveleek can help with this better than I can

gastonfournier (commented):
LG from my point of view, but I'd love to have David's input. That `null!` is not something I've seen in Java

dotnet-engine/Yggdrasil.Engine/Types.cs
daveleek: The other alternative is to use constructors with non-nullable parameters - which should cause exceptions on deserialization

dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -13,29 +13,31 @@ public class Context
...
 }
 
-public class Variant
+public sealed class Variant
 {
-    public string Name { get; set; }
+    public string Name { get; set; } = null!;
     public Payload? Payload { get; set; }
-    public bool Enabled { get; set; }
+    public bool Enabled { get; set; } = false;
```
daveleek, line 34: This is interesting - it shouldn't be needed. bools default to false

daveleek (approved):
LG, shame we have a bit of double effort, lets have a chat on how to merge this

sighphyre:
Think this one got superseded, going to close this

PR #73, opened 2023-11-08
chore: basic isEnabled benchmarks

sighphyre:
Adds some benchmarks purely for the equivalent `isEnabled` call for Ruby, Java and .NET. These are not intended to be extensive benchmarks, just enough to check that each SDK is correctly managing the FFI layer reasonably and get a loose sense of the computational differences between the implementations.

Each implementation has instructions in the relevant README for running the benchmarks.

Output converted to a human readable, comparable form:

Java - 123 k operations/s 
Ruby - 283 k operations/s
.NET - 493 k operations/s

Ruby is roughly twice the speed of the equivalent code in the SDK, which is within a reasonable target. Haven't checked how .net compares to the baseline implementation. Something appears to be off with Java, but that can be discussed elsewhere.

### DIFF
--- a/dotnet-engine/Yggdrasil.Benchmarks/Program.cs
+++ b/dotnet-engine/Yggdrasil.Benchmarks/Program.cs
@@ -1,0 +1,60 @@
+﻿using System;
+using System.Text.Json;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Running;
+using Newtonsoft.Json.Linq;
+using Yggdrasil;
+
+[Config(typeof(Config))]
+public class YggBench
+{
+    private YggdrasilEngine yggdrasilEngine;
+
+    private class Config : ManualConfig
+    {
+        public Config()
+        {
+            AddColumn(BenchmarkDotNet.Columns.StatisticColumn.OperationsPerSecond);
+        }
+    }
+
+    [GlobalSetup]
+    public void Setup()
+    {
+        var basePath = Path.Combine(
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "client-specification",
+            "specifications"
+        );
+        var suitePath = Path.Combine(basePath, "01-simple-examples.json");
+        var suiteData = JObject.Parse(File.ReadAllText(suitePath));
+
+        yggdrasilEngine = new YggdrasilEngine();
+        yggdrasilEngine.TakeState(suiteData["state"].ToString());
+    }
+
+    [Benchmark]
+    public void IsFeatureAEnabled()
+    {
+        yggdrasilEngine.IsEnabled("Feature.A", new Context());
+    }
+}
+
+public class Program
+{
+    public static void Main(string[] args)
+    {
+        var summary = BenchmarkRunner.Run<YggBench>();
+    }
+}
--- a/java-engine/build.gradle
+++ b/java-engine/build.gradle
@@ -1,6 +1,7 @@
 plugins {
     // Apply the java-library plugin for API and implementation separation.
     id 'java-library'
+    id "me.champeau.jmh" version "0.7.2"
 }
 
 repositories {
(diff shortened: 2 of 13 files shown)


dotnet-engine/Yggdrasil.Benchmarks/Program.cs
```diff
@@ -0,0 +1,60 @@
...
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "..",
+            "client-specification",
```
sighphyre, line 37: @daveleek I feel like I've downed a bottle of crazy pills here. Where the hell is this assembly actually executing!?

Surely there's something better than this crazy down path?

dotnet-engine/Yggdrasil.Benchmarks/Program.cs
```diff
@@ -0,0 +1,60 @@
...
+[Config(typeof(Config))]
+public class YggBench
+{
+    private YggdrasilEngine yggdrasilEngine;
+
+    private class Config : ManualConfig
+    {
+        public Config()
+        {
+            AddColumn(BenchmarkDotNet.Columns.StatisticColumn.OperationsPerSecond);
```
sighphyre, line 20: Not strictly necessary but gives us the output we want to be able to compare against the others

java-engine/settings.gradle
```diff
@@ -6,6 +6,12 @@
  * Detailed information about configuring a multi-project build in Gradle can be found
  * in the user manual at https://docs.gradle.org/8.1/userguide/multi_project_builds.html
  */
+pluginManagement {
```
sighphyre, line 9: No idea why this needed to happen but Gradle flat out refused to resolve jmh without it for some reason

java-engine/build.gradle
```diff
@@ -35,3 +40,14 @@ tasks.named('test') {
     // Use JUnit Platform for unit tests.
     useJUnitPlatform()
 }
+
+jmh {
```
sighphyre, line 44: The default settings for jmh spin off a ~30 minute benchmark run. I do not have the patience for that so I've fiddled so that they execute in a similar 30-60 off seconds like the others but this config could use more eyes.

@gastonfournier sanity check?

ruby-engine/README.md
```diff
@@ -15,10 +15,18 @@ Then you can run the tests with:
...
+rspec scripts/benchmark.rb
+```
+And should produce human readable output.
+
+
 There's also a `mem_check.rb` in the scripts folder. This is not a bullet proof test, but it can be helpful for detecting large leaks. This requires human interaction - you need to read the output and understand what it's telling you, so it's not run as part of the test suite.
 
 ```bash
-ruby scripts/mem_check.rb
+rspec scripts/mem_check.rb
```
sighphyre, line 29: Just boyscouting, the previous command never worked

java-engine/src/jmh/java/UnleashEngineBenchmark.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private static String loadFeaturesFromFile(String filePath) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        JsonNode jsonNode = mapper.readTree(Paths.get(filePath).toFile());
+        JsonNode state = jsonNode.get("state");
+        return state.toString();
+    }
+
+    @Setup
+    public void setUp() {
```
gardleopard, line 28: try setting the context here and reuse it in the benchmarkFeatureToggle() method

gardleopard (approved):
LGTM

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -77,7 +77,6 @@ private <T> T read(Pointer pointer, TypeReference<T> typeReference) {
         String str = pointer.getString(0, UTF_8);
         yggdrasil.freeResponse(pointer);
         try {
-            System.out.println(str); // TODO use a logging library. SLF4J?
```
sighphyre, line 80: Yeeted this for now since std in a loop generally hurts benchmark reliability

dotnet-engine/Yggdrasil.Benchmarks/Program.cs
daveleek: Long time since I looked at that, but back in the day mstest would move things to run in a subfolder deep down. you could output `TestContext.CurrentContext.TestDirectory` to the console and have a look 😅

daveleek (approved):
.NET stuff looks good to me!

dotnet-engine/Yggdrasil.Benchmarks/Program.cs
sighphyre: Don't think that will work, this isn't running in a test context, its a standalone executable so I can remove test harness fluff from the benchmarks

But if nothing stands out to you, it can sit, it works, it's just ugly

java-engine/src/jmh/java/UnleashEngineBenchmark.java
sighphyre: Yep! Gave it a spin, no meaningful change to output

gastonfournier (approved):
LG!

PR #77, opened 2023-11-09
feat: yggdrasil dotnet custom strategies

daveleek:
## About the changes

Implements custom strategy handling in Yggdrasil .NET.
Adds some tests
Improves build of Yggdrasil.NET Engine


dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -67,3 +67,76 @@ public class MetricsBucket
...
+{
+    public List<Feature>? Features { get; set; }
+}
+
+class MappedFeature
+{
+    public MappedFeature(Feature feature, List<MappedStrategy> strategies)
+    {
+        Name = feature.Name;
+        Strategies = strategies;
```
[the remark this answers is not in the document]
daveleek: That's a very good point, I'll see if I can provoke the behavior a bit with some unit tests!

dotnet-engine/Yggdrasil.Engine.Tests/Utils.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Yggdrasil.Test;
+
+public class CustomStrategyReturningTrue : IStrategy
+{
+    public CustomStrategyReturningTrue(string name)
+    {
+        Name = name;
+    }
+
+    public string Name { get; private set; }
```
sighphyre, line 12: Nit: Can we move this to the top so we can have methods together and fields together?

dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
```diff
@@ -14,10 +15,10 @@ public class Tests
         PropertyNamingPolicy = JsonNamingPolicy.CamelCase
     };
 
-    [Test]
+    //[Test]
```
sighphyre, line 18: What's the plan here?

If we want to keep this, let's leave it in. If it needs refactor to make this easier and saner to use, let's do that in another PR

```suggestion
    [Test]

```

dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
```diff
@@ -100,6 +102,54 @@ public void TestClientSpec()
         }
     }
 
+    [Test]
+    public void Custom_Strategies_Required_But_Not_Configured_Fails() {
```
sighphyre, line 106: I don't think this is a fail, this should degrade gracefully

```suggestion
    public void Custom_Strategies_Required_But_Not_Configured_Returns_False() {

```

dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -27,7 +27,7 @@ public class EngineResponse<TValue> : EngineResponse {
 
 public class Variant
 {
-    public string Name { get; set; }
+    public string? Name { get; set; }
```
sighphyre, line 30: This looks a little suspect to my eyes, this should never be null, right?

dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -45,6 +45,11 @@ public override bool Equals(object? obj)
         var payload = (Payload)obj;
         return Value == payload.Value && PayloadType == payload.PayloadType;
     }
+
+    public override int GetHashCode()
+    {
+        return HashCode.Combine(Value, PayloadType);
```
sighphyre, line 51: Cool! That's much better than what I wanted to do!

dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -56,14 +61,84 @@ public class FeatureCount
...
     public long No { get; set; }
-    public Dictionary<string, long> Variants { get; set; }
+    public Dictionary<string, long>? Variants { get; set; }
 }
 
 
 public class MetricsBucket
 {
-    public Dictionary<string, FeatureCount> Toggles { get; set; }
+    public Dictionary<string, FeatureCount>? Toggles { get; set; }
```
sighphyre, line 70: Also doesn't look right to me, this should be set if this object exists

dotnet-engine/Yggdrasil.Engine/YggdrasilEngine.cs
```diff
@@ -17,9 +19,34 @@ public YggdrasilEngine()
...
+    public void RegisterCustomStrategies(List<IStrategy> strategies)
+    {
+        customStrategies.RegisterCustomStrategies(strategies);
+    }
+
     public bool ShouldEmitImpressionEvent(string featureName)
     {
-        return FFI.ShouldEmitImpressionEvent(state, featureName);
+        var shouldEmitImpressionEventPtr = FFI.ShouldEmitImpressionEvent(state, featureName);
+        if (shouldEmitImpressionEventPtr == IntPtr.Zero)
```
sighphyre, line 30: Okay, I see this is a pattern we're using elsewhere but this needs to change in another PR. Nil pointers are a serious problem since the underlying FFI layer can't return those. So something very, very wrong has happened and we need to know about it

I want to say we should draw inspiration from the Java engine here and have a single method can unwrap the incoming FFI JSON bodies and deal with them

sighphyre (changes requested):
The underlying algorithm looks correct to me but there's a bunch of new warnings and some suspect nullable types that I would like to resolve before we merge this

dotnet-engine/Yggdrasil.Engine.Tests/custom-strategies.json
```diff
@@ -0,0 +1,49 @@
+{
```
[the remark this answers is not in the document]
gastonfournier: Yes, I already did the custom-strategies-test inside Java, I could benefit from having this one merged

dotnet-engine/Yggdrasil.Engine.Tests/Utils.cs
daveleek: Done!

dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
daveleek: The plan is to ensure that our builds run quickly while we look at a good way to separate tests to be run at build and tests to be run when explicitly invoked, and that seems to be test categories but I wanted to save that for a new PR when I've had a chance to try it out properly

dotnet-engine/Yggdrasil.Engine/YggdrasilEngine.cs
daveleek: Yeah I agree, and yeah lets open a new PR with that fix

dotnet-engine/Yggdrasil.Engine/YggdrasilEngine.cs
```diff
@@ -30,6 +57,8 @@ public void Dispose()
 
     public void TakeState(string json)
     {
+        customStrategies.MapFeatures(json);
```
gastonfournier, line 60: I think you should move this after taking the state and error handling.

daveleek: Good point!

dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
sighphyre: Can we remove the commented out stuff then?

Not gonna block this for that though

sighphyre (approved):
G

PR #80, opened 2023-11-13
fix: make free floating quotes in string operators work correctly

sighphyre:
This makes the rule upgrader escape incoming string operators if they contain free floating quotes. So this:

`user_id in [" look me Im "cool"  "]`
Should escape to :
`user_id in [" look me Im \"cool\"  "]`

This then gets removed during evaluation phase so that we don't have to escape quotes in the context


nunogois (approved):
LGTM!

PR #81, opened 2023-11-13
chore(java): improve memory usage

gastonfournier:
Make some changes in UnleashEngine internals to release memory properly

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
+++ b/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
@@ -4,8 +4,6 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.ObjectReader;
-import com.fasterxml.jackson.databind.ObjectWriter;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.sun.jna.Pointer;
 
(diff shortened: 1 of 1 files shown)


java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -66,21 +63,25 @@ public void countVariant(String flagName, String variantName) {
...
         try {
-            System.out.println(str); // TODO use a logging library. SLF4J?
-            return reader.forType(typeReference).readValue(str);
-        } catch (IOException e) {
-            throw new YggdrasilParseException(str, typeReference.getClass(), e);
+            String str = pointer.getString(0, UTF_8);
+            try {
+                return mapper.readValue(str, clazz);
+            } catch (IOException e) {
+                System.out.println("Failed to parse response from Yggdrasil: " + str);
```
sighphyre, line 80: Can we trim this? Certainly for benchmarks, streaming to std out is costly. I removed this line when I ran the original benchmarks
```suggestion

```

gastonfournier: Yes, I'd like to integrate with SLF4J and add this as debug log, but I'd like to keep it for now to later replace all System.out with the logger

sighphyre: Fair enough, but you'll need to trim this out for your benchmark runs to get anything honest

sighphyre (approved):
G

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
gastonfournier: Yes, the logger will just be disabled for benchmarks

PR #83, opened 2023-11-13
chore: build dependents on ygg

gastonfournier:
When Yggdrasil or the FFI changes, build the dependent sub-projects. This will give us a fast feedback loop without having to build them in every PR


sighphyre (approved):
Nice!

PR #84, opened 2023-11-14
fix/feature parity with ruby sdk

sighphyre:
This adds a method to resolve variants to a compatible type for the Ruby SDK. Previously our methods would return a variant object that had everything mapped as symbols (this is good). The Ruby SDK expects a mix of symbols and strings. I'd love to change this but this will likely break existing code in the wild that's in use


ruby-engine/ext/extconf.rb
```diff
@@ -0,0 +1,2 @@
+require 'mkmf'
+create_makefile('unleash-engine/unleash-engine')
```
sighphyre, line 2: Snuck in because I need a useful build for ruby and this doesn't work without this. Plz ignore, soz

gastonfournier (approved):
LGTM!

PR #85, opened 2023-11-14
chore: expose known strategies

gastonfournier:
This is to be used in host languages to know if a strategy is implemented inside unleash-yggdrasil or requires a custom strategy implementation on the host language


sighphyre (approved):
G

PR #86, opened 2023-11-15
chore(java): custom strategies implementation

gastonfournier:
- Custom strategies implementation and testing
- SLF4J API for logging without an implementation (if available in the classpath it should pick it up and output logs)


java-engine/src/test/resources/impression-data-tests.json
```diff
@@ -17,6 +17,13 @@
       "strategies": [{
         "name": "default"
       }]
+    },
+    {
+      "name": "with.impression.data.undefined",
+      "enabled": true,
+      "strategies": [{
+        "name": "default"
+      }]
```
gastonfournier, line 26: @daveleek an additional case

java-engine/src/test/resources/custom-strategy-tests.json
```diff
@@ -0,0 +1,87 @@
+{
+  "version": 2,
+  "features": [
+    {
+      "name": "Feature.Custom.Strategies",
```
gastonfournier, line 5: This file is the same as .Net, we should move it to a common place.

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -7,39 +7,77 @@
...
+            this.customStrategies =
+                    new CustomStrategies(
+                            customStrategies.stream()
+                                    .filter(
+                                            strategy -> {
+                                                if (builtInStrategies.contains(
+                                                        strategy.getName())) {
+                                                    log.warn(
+                                                            "Custom strategy {} is already a built-in strategy. Skipping.",
+                                                            strategy.getName());
```
sighphyre, line 54: This indentation tho. Is this normal? This is what the formatter wants?

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -88,7 +126,7 @@ private <T> T read(Pointer pointer, TypeReference<T> clazz) {
             try {
                 return mapper.readValue(str, clazz);
             } catch (IOException e) {
-                System.out.println("Failed to parse response from Yggdrasil: " + str);
+                log.error("Failed to parse response from Yggdrasil: {}", str, e);
```
sighphyre, line 129: ❤️

java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
```diff
@@ -44,6 +49,43 @@ public static String loadFeaturesFromFile(String filePath) {
         }
     }
 
+    static IStrategy alwaysTrue =
```
sighphyre, line 52: Wow, I've been away from Java a long time, I didn't even know you could do this now!

java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
```diff
@@ -197,45 +239,81 @@ void testMetrics() throws YggdrasilError {
...
+    @ParameterizedTest
+    @MethodSource("customStrategiesInput")
+    void customStrategiesRequired_whenNotConfigured_returnsFalse(
+            List<IStrategy> customStrategies,
+            String featureName,
+            Context context,
+            boolean expectedIsEnabled)
+            throws Exception {
+        UnleashEngine customEngine =
+                new UnleashEngine(new YggdrasilFFI("../target/release"), customStrategies);
```
sighphyre, line 268: Next thing to poke at I think. We should get away from yeeting native paths at the engine soonish

java-engine/src/main/java/io/getunleash/engine/CustomStrategies.java
```diff
@@ -0,0 +1,151 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class CustomStrategies {
```
sighphyre, line 18: This name is a little off to my eyes. This doesn't tell me a lot about what it does. It especially looks weird for a class called CustomStrategies to take a parameter called customStrategies in its constructor.

CustomStrategyHandler? Dunno if I like that but it tells me more about what this does

java-engine/src/main/java/io/getunleash/engine/CustomStrategies.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private Map<String, List<MappedStrategy>> featureStrategies = new HashMap<>();
+
+    public CustomStrategies(Stream<IStrategy> customStrategies) {
+        this.mapper = new ObjectMapper();
+        this.mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        this.registeredStrategies = customStrategies.collect(toMap(IStrategy::getName, identity()));
+    }
+
+    public void takeState(String toggles) {
```
sighphyre, line 32: Also think this name could use some love. updateCustomStrategies maybe? It's too close to the Yggdrasil core takeState to my eyes

sighphyre (approved):
Nice! I've left some suggestions but I think they're optional

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
gastonfournier: Yeah, I don't like it either. I'm going to look into more detailed configuration of spotless to make it look better. I just took the defaults for now

java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
gastonfournier: Oh, those constructors are not public, they're just for tests. The public ones: https://github.com/Unleash/yggdrasil/blob/5eabc28e6d6c55aea6b293605aae006fa3bda9e2/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java#L25-L31

java-engine/src/main/java/io/getunleash/engine/CustomStrategies.java
gastonfournier: Yes, I carried this over from .Net, I was unsure about this name. I like your proposal but let me do some thinking before stamping that name :D

java-engine/src/main/java/io/getunleash/engine/CustomStrategies.java
gastonfournier: Yes, it was intentional, because it happens on take state. I don't like that CustomStrategies now have to know about the shape of toggles input, because that's duplicated in Yggdrasil and will be replicated in all Engines... I was thinking how to solve this problem but we should probably talk about this

java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
sighphyre: Okay yeah that's fair!

java-engine/src/main/java/io/getunleash/engine/CustomStrategies.java
sighphyre: Yeah I hate it too. I don't have a better suggestion right now though. My ideal scenario is we throw away custom strategies which will make this go away

This is basically a hack to get us over the line and put us in a place where we can do that

PR #87, opened 2023-11-15
feat: add warnings as errors in tests proj and fix context possibly b…

daveleek:
add warnings as errors in tests proj as well and fix context possibly being null


sighphyre (approved):
Nice!

PR #88, opened 2023-11-15
chore: rename everything to Yggrrasil rather than Unleash in ruby engine

sighphyre:
Big rename of all the ruby implementations from UnleashEngine to Yggdrasil engine, to keep it inline with .NET and our naming strategy


gastonfournier (approved):
LG!

PR #90, opened 2023-11-15
chore(java): java 8 support

gastonfournier:
This adds Java 8 target to be able to import the engine in Java 8 projects


sighphyre (approved):
Nice, simpler than I expected!

PR #91, opened 2023-11-15
feat: add feature_enabled to variant response

nunogois:
https://linear.app/unleash/issue/SR-158/yggdrasil-edge-add-feature-enabled-property-to-variant-response

Bumps the client spec and adds a new `feature_enabled` property to the variant response, similar to what we've done for other SDKs.

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -325,7 +325,7 @@ impl EngineState {
 
             let parent_enabled = self.enabled(compiled_parent, context, &None); //parent toggles explicitly don't support custom strategies
             let expected_parent_enabled_state = parent_dependency.enabled.unwrap_or(true);
-            let parent_variant = self.check_variant_by_toggle(compiled_parent, context);
+            let parent_variant = self.check_variant_by_toggle(compiled_parent, context, parent_enabled);
 
             let is_variant_dependency_satisfied = {
                 if let (Some(expected_variants), Some(actual_variant)) =
@@ -471,6 +471,7 @@ impl EngineState {
         &self,
         toggle: &CompiledToggle,
         context: &Context,
+        is_enabled: bool,
     ) -> Option<VariantDef> {
         let strategy_variants =
             toggle
@@ -498,7 +499,7 @@ impl EngineState {
             name: variant.name.clone(),
             payload: variant.payload.clone(),
             enabled: true,
-            feature_enabled: true,
+            feature_enabled: is_enabled.clone(),
         })
     }
 
@@ -510,7 +511,7 @@ impl EngineState {
     ) -> Option<VariantDef> {
         self.get_toggle(name).and_then(|toggle| {
             if self.enabled(toggle, context, external_values) {
-                self.check_variant_by_toggle(toggle, context)
+                self.check_variant_by_toggle(toggle, context, true)
             } else {
                 None
             }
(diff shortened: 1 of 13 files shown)


gastonfournier (approved):
Java looks good

java-engine/src/main/java/io/getunleash/engine/YggdrasilFFI.java
```diff
@@ -62,7 +62,7 @@ static UnleashFFI loadLibrary(String libraryPath) {
             libImpl = "libyggdrasilffi.dll";
         }
 
-        String combinedPath = Paths.get(libraryPath, libImpl).toString();
+        String combinedPath = Paths.get(libraryPath, libImpl).toAbsolutePath().toString();
```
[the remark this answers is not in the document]
nunogois: I think it makes sense as a separate PR like we've mentioned, so I'll do that.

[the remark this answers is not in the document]
nunogois: https://github.com/Unleash/yggdrasil/pull/92

unleash-yggdrasil/src/lib.rs
```diff
@@ -529,7 +530,10 @@ impl EngineState {
             self.count_toggle(name, enabled);
 
             if enabled {
-                self.check_variant_by_toggle(toggle, context)
+                let mut variant = self.check_variant_by_toggle(toggle, context).unwrap_or_default();
+                variant.feature_enabled = true;
+                println!("Setting feature_enabled");
```
sighphyre, line 535: Yeet!
```suggestion
```

unleash-yggdrasil/src/lib.rs
```diff
@@ -540,6 +544,7 @@ impl EngineState {
         .unwrap_or_default();
 
         self.count_variant(name, &variant.name);
+        println!("Returning variant with feature_enabled: {}", &variant.feature_enabled);
```
sighphyre, line 547: Yeet!
```suggestion
```

unleash-yggdrasil/src/lib.rs
```diff
@@ -529,7 +530,9 @@ impl EngineState {
             self.count_toggle(name, enabled);
 
             if enabled {
-                self.check_variant_by_toggle(toggle, context)
+                let mut variant = self.check_variant_by_toggle(toggle, context).unwrap_or_default();
+                variant.feature_enabled = true;
```
sighphyre, line 534: I don't think this is quite going to work for us here. My feeling is that check_variant_by_toggle should be told what the state of the parent toggle is rather than setting a potentially invalid default and hoping the caller fixes it.

From a quick glance, it looks like everywhere that `check_variant_by_toggle` is called, we have enough information to pass that in. Then it would be much tougher to make a mistake on this

Can we change the function signature to `check_variant_by_toggle(toggle, context, is_enabled)`?

unleash-yggdrasil/src/lib.rs
```diff
@@ -612,6 +615,7 @@ pub struct VariantDef {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub payload: Option<Payload>,
     pub enabled: bool,
+    pub feature_enabled: bool,
 }
 
 impl Default for VariantDef {
```
sighphyre, line 621: I think the default impl here is only used for tests but it's now definitely dangerous to use. Can we move this into the test mod so that it's no longer exposed

sighphyre (changes requested):
This is pretty great! Thank you for tackling literally everything - I had only expected the Rust patch here

One or two places where I do think we should make a change though

daveleek (approved):
.NET Looks good! Thanks @nunogois

PR #92, opened 2023-11-15
fix: use absolutepath, fixing path in macos

nunogois:
Fixes the path in MacOS, like mentioned in https://github.com/Unleash/yggdrasil/pull/91#discussion_r1394103945


gastonfournier (approved):
Go go!

PR #93, opened 2023-11-15
style: formatting

nunogois:
Seems like my formatter disagrees with some of the current formatting. I think it's probably right, since this feels more consistent to me.


sighphyre (approved):
G

PR #95, opened 2023-11-16
chore(dotnet): standardize formatting on dotnet format

sighphyre:
Apply formatting, put in a standard, be done with this


daveleek (approved):
YES!

PR #96, opened 2023-11-16
chore(java): handling of fallback strategy

gastonfournier:
## About the changes
After doing a first integration in Java SDK these are feature that were missing in the Engine

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
+++ b/java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
@@ -29,7 +29,8 @@ class CustomStrategiesEvaluator {
         this(customStrategies, null);
     }
 
-    public CustomStrategiesEvaluator(Stream<IStrategy> customStrategies, IStrategy fallbackStrategy) {
+    public CustomStrategiesEvaluator(
+            Stream<IStrategy> customStrategies, IStrategy fallbackStrategy) {
         this.mapper = new ObjectMapper();
         this.mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
         this.registeredStrategies =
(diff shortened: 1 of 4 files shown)


java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
```diff
@@ -65,14 +73,20 @@ private List<MappedStrategy> getFeatureStrategies(FeatureDefinition feature) {
             mappedStrategies.add(
                     new MappedStrategy("customStrategy" + (index++), impl, strategyDefinition));
         }
+        if (fallbackStrategy != null) {
+            mappedStrategies.add(
+                    new MappedStrategy(
+                            "customStrategy" + index,
+                            fallbackStrategy,
+                            new StrategyDefinition("fallback", Collections.emptyMap())));
```
gastonfournier, line 81: The name doesn't really matter, this strategy has to be available for all feature toggles

java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
```diff
@@ -100,6 +101,7 @@ void testGetVariant() throws Exception {
     @Test
     public void testClientSpec() throws Exception {
         ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
gastonfournier, line 104: Just for easier evolution: don't fail when new fields are added.

sighphyre:
I'm not convinced this is the correct thing to do here. This looks like a very odd feature in the Java SDK. Can we move towards deprecating that rather?

gastonfournier:
> I'm not convinced this is the correct thing to do here. This looks like a very odd feature in the Java SDK. Can we move towards deprecating that rather?

I can mark it as deprecated in the engine's public interface and then in the SDK, but right now this is needed for backward compatibility

sighphyre (approved):
G

PR #97, opened 2023-11-17
feat(yggdrasil): Bubble up errors on take state

gastonfournier:
When taking state, different types of parsing errors can happen. In such situations, Yggdrasil should fail gracefully communicate the error to the client.


sighphyre (approved):
LGTM

PR #98, opened 2023-11-23
feat: add feature enabled property to variant checks 

sighphyre:
This add the enabled status of the toggle to the response in get_variant and does effectively the same logic in the FFI layers. This is a redesign of https://github.com/Unleash/yggdrasil/pull/91, to make the variant handling code a little less invasive and a little safer. This is the important PR since it allows us to do this in Edge.

Notably this does not handle the new property in the `check_variant` method in the core layer of Yggdrasil, instead it makes this the FFI layers problem, which I think is okay, since the FFI code has all the information it needs to be able to determine this. This is encoded in the type signatures of the respective methods so it should be tough to misuse.

Handling code for Java, Ruby and Dotnet is in other PRs to make this easier to review. Patches for the incomplete engines has been abandoned since they need a rewrite anyway. None of the FFI engines are in production so it should be safe to split

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -28,7 +28,7 @@ use unleash_types::client_metrics::{MetricBucket, ToggleStats};
 
 pub type CompiledState = HashMap<String, CompiledToggle>;
 
-pub const SUPPORTED_SPEC_VERSION: &str = "5.0.2";
+pub const SUPPORTED_SPEC_VERSION: &str = "5.1.0";
 const VARIANT_NORMALIZATION_SEED: u32 = 86028157;
 
 pub struct CompiledToggle {
@@ -208,7 +208,7 @@ pub struct ResolvedToggle {
     pub enabled: bool,
     pub impression_data: bool,
     pub project: String,
-    pub variant: VariantDef,
+    pub variant: ExtendedVariantDef,
 }
 
 impl EngineState {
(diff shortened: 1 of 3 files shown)


unleash-yggdrasil/src/lib.rs
```diff
@@ -528,26 +528,22 @@ impl EngineState {
...
-
-            if enabled {
-                self.check_variant_by_toggle(toggle, context)
-            } else {
-                None
-            }
-        } else {
-            self.count_toggle(name, false);
-            None
+        let variant = match toggle {
```
sighphyre, line 537: Does the same as the block it removes,(minus the count which is now handled after this block).

This was getting a bit ugly and since we're in this space...

unleash-yggdrasil/src/lib.rs
```diff
@@ -622,6 +618,26 @@ pub struct VariantDef {
...
+            name: self.name.clone(),
+            payload: self.payload.clone(),
+            enabled: self.enabled,
+            feature_enabled: toggle_enabled,
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Clone)]
+pub struct ExtendedVariantDef {
```
sighphyre, line 633: Honestly not loving this name. If anyone has a better one please feel free to suggest

yggdrasilffi/src/lib.rs
```diff
@@ -211,14 +211,21 @@ pub unsafe extern "C" fn check_variant(
...
         let custom_strategy_results =
             get_json::<CustomStrategyResults>(custom_strategy_results_ptr)?;
-
-        Ok(engine.check_variant(toggle_name, &context, &Some(custom_strategy_results)))
+        let base_variant = engine.check_variant(
+            toggle_name,
+            &context,
+            &Some(custom_strategy_results.clone()),
+        );
+        let toggle_enabled = engine
```
sighphyre, line 225: We'll be checking this twice since the consumer will have to anyway but the code that does the check is literally several orders of magnitude faster than moving across the FFI boundary so I think this is okay from a perf perspective

unleash-yggdrasil/src/lib.rs
nunogois: Same. Honestly I see this simply as a VariantDef, since this is how our variants now look like, with the `feature_enabled` property as part of them. So we're saying a variant is now effectively:

```
{
  name,
  enabled,
  payload,
  feature_enabled
}
```

Which means your `variant` variable is not really a variant, and `check_variant_by_toggle` doesn't really return a variant, but rather parts of the variant.

Perhaps something like:
 - `VariantDef` -> `BaseVariantDef`
 - `ExtendedVariantDef` -> `VariantDef`
 
 But to be honest I'm not 100% sure. I think this needs more input.

nunogois (approved):
Looks sane to me. Not sure how to feel about the `VariantDef` / `ExtendedVariantDef` dynamic, but perhaps they just need better naming.

If you're happy then I'm happy 😄

PR #100, opened 2023-11-23
feat: allow feature enabled property to propagate in java engine

sighphyre:
Depends on https://github.com/Unleash/yggdrasil/pull/98

Basically just allows that to propagate out. Shameless stolen from https://github.com/Unleash/yggdrasil/pull/91

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/VariantDef.java
+++ b/java-engine/src/main/java/io/getunleash/engine/VariantDef.java
@@ -2,20 +2,24 @@ package io.getunleash.engine;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Optional;
 
 public class VariantDef {
     private final String name;
     private final Payload payload;
     private final Boolean enabled;
+    private final Boolean featureEnabled;
 
     @JsonCreator
     VariantDef(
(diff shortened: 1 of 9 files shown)


java-engine/src/main/java/io/getunleash/engine/VariantDef.java
```diff
@@ -2,20 +2,24 @@
...
     VariantDef(
             @JsonProperty("name") String name,
             @JsonProperty("payload") Payload payload,
-            @JsonProperty("enabled") Boolean enabled) {
+            @JsonProperty("enabled") Boolean enabled,
+            @JsonProperty("feature_enabled") Boolean featureEnabled) {
         this.name = name;
         this.payload = payload;
         this.enabled = enabled;
+        this.featureEnabled = Optional.ofNullable(featureEnabled).orElse(false);
```
gastonfournier, line 22: I still find it confusing to have enabled and feature enabled as part of the variant def, but I think I'm late to this conversation

gastonfournier (approved):
LGTM!

java-engine/src/main/java/io/getunleash/engine/VariantDef.java
sighphyre: I don't disagree but I do think you're a bit late to the party. This is already in the client spec and implemented in Python

PR #101, opened 2023-11-24
feat(core): implement hostname strategy in core

sighphyre:
This adds a hostname strategy implementation for Yggdrasil core.

### DIFF
--- a/unleash-yggdrasil/src/strategy_grammar.pest
+++ b/unleash-yggdrasil/src/strategy_grammar.pest
@@ -64,6 +64,7 @@ constraint = {
     invert_operation* ~
     (
         rollout_constraint
+        | hostname_constraint
         | date_constraint
         | semver_constraint
         | numeric_constraint
@@ -74,6 +75,9 @@ constraint = {
     )
 }
     default_strategy_constraint = { boolean }
+    hostname_constraint = { hostname ~ in ~ string_list }
+        hostname = _{ "hostname" }
+        in = _{ "in" }
     string_fragment_constraint = { context_value ~ ( string_list_operation_without_case | string_list_operation ) ~ string_list }
     list_constraint = { context_value ~ list_operation ~ ( numeric_list | string_list | empty_list ) }
     date_constraint = { context_value ~ ordinal_operation ~ date }
--- a/unleash-yggdrasil/src/strategy_parsing.rs
+++ b/unleash-yggdrasil/src/strategy_parsing.rs
@@ -1,12 +1,15 @@
 extern crate pest;
 
 use std::collections::HashSet;
+use std::env;
 use std::io::Cursor;
 use std::num::ParseFloatError;
 
 use crate::sendable_closures::{SendableContextResolver, SendableFragment};
+use crate::state::SdkError;
 use crate::EnrichedContext as Context;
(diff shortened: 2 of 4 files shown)


unleash-yggdrasil/src/strategy_grammar.pest
```diff
@@ -74,6 +75,9 @@ constraint = {
     )
 }
     default_strategy_constraint = { boolean }
+    hostname_constraint = { hostname ~ in ~ string_list }
+        hostname = _{ "hostname" }
```
sighphyre, line 79: The little underscore here is a non emitting rule - it doesn't generate a node to iterate over in the parser. Chosen this because we don't care about anything outside of the contents of the list

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -327,6 +330,30 @@ fn rollout_constraint(mut node: Pairs<Rule>) -> RuleFragment {
...
+                .map_err(|_| SdkError::StrategyEvaluationError)
+        })
+}
+
+fn hostname_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
+    let target_hostnames = harvest_set(node.next().unwrap().into_inner());
+
+    Box::new(move |_: &Context| match get_hostname() {
+        Ok(hostname) => target_hostnames.contains(&hostname).invert(inverted),
+        Err(_) => false,
```
sighphyre, line 353: I've had to make a decision here because I don't see any other SDK outside of Rust handling this error case (this is an OS call or env var read so this _can_ fail). This would disable the strategy if retrieving the hostname fails, means means the strategy would be ignored

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -327,6 +330,30 @@ fn rollout_constraint(mut node: Pairs<Rule>) -> RuleFragment {
     })
 }
 
+fn get_hostname() -> Result<String, SdkError> {
+    //This is primarily for testing purposes
+    if let Ok(hostname_env) = env::var("HOSTNAME") {
```
sighphyre, line 335: Hard to do an automated test in any other way since the hostname crate does an OS call rather than an env var read. Validated locally against my own machine though

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -327,6 +330,30 @@ fn rollout_constraint(mut node: Pairs<Rule>) -> RuleFragment {
...
+        .map_err(|_| SdkError::StrategyEvaluationError)
+        .and_then(|os_str| {
+            os_str
+                .into_string()
+                .map_err(|_| SdkError::StrategyEvaluationError)
+        })
+}
+
+fn hostname_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
+    let target_hostnames = harvest_set(node.next().unwrap().into_inner());
```
sighphyre, line 349: This is the magic referenced earlier - we only have to read a single node rather than all 3 nodes since the first two nodes are non emitting. They tell us nothing so they're discarded in the grammar rather than the compiler

gastonfournier (approved):
LGTM!

PR #102, opened 2023-11-24
fix(core): strategies that are broken will now default to false

sighphyre:
This allows the code path that converts raw strategy structs to the DSL format to fallback to "false" when a recoverable error is encountered. Typically this is because no parameters were passed for the strategy. This means the strategies will be ignored when evaluated.

This isn't necessary to pass the spec tests and "should" never happen but life doesn't work like that and we can and should handle recoverable errors.

### DIFF
--- a/unleash-yggdrasil/src/strategy_upgrade.rs
+++ b/unleash-yggdrasil/src/strategy_upgrade.rs
@@ -158,11 +158,13 @@ fn upgrade_strategy(
 }
 
 fn upgrade_flexible_rollout_strategy(strategy: &Strategy) -> String {
-    let rollout = strategy.get_param("rollout");
+    let rollout = strategy
+        .get_param("rollout")
+        .map(|value| value.parse::<usize>())
+        .and_then(Result::ok);
+
     match rollout {
         Some(rollout) => {
-            //should probably validate at this point that the rollout looks like a percent
-
             let mut rule: String = format!("{rollout}%");
 
             rule = format!(
@@ -190,7 +192,7 @@ fn upgrade_user_id_strategy(strategy: &Strategy) -> String {
                 .join(",");
             format!("user_id in [{user_ids}]")
         }
-        None => "".into(),
+        None => "false".into(),
     }
 }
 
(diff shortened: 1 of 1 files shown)


unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -158,11 +158,13 @@ fn upgrade_strategy(
 }
 
 fn upgrade_flexible_rollout_strategy(strategy: &Strategy) -> String {
-    let rollout = strategy.get_param("rollout");
+    let rollout = strategy
```
sighphyre, line 161: Validates that our rollout parameters are actually a numeric type. This gets its own code because in flexible rollout we use a completely different parameter name

unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -207,39 +209,50 @@ fn upgrade_remote_address(strategy: &Strategy) -> String {
...
+    let percentage = get_rollout_percent(strategy);
+
+    match percentage {
         Some(percent) => format!("random < {percent}"),
-        None => "".into(),
+        None => "false".into(),
     }
 }
 
+fn get_rollout_percent(strategy: &Strategy) -> Option<usize> {
```
sighphyre, line 249: Utility method

unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -821,4 +834,53 @@ mod tests {
...
     }
+
+    #[test_case("gradualRolloutUserId")]
+    #[test_case("gradualRolloutSessionId")]
+    #[test_case("gradualRolloutRandom")]
+    fn gradual_rollout_strategies_with_invalid_parameters_are_false(strategy_type: &str) {
+        let strategy = Strategy {
+            name: strategy_type.into(),
+            parameters: Some(
+                vec![("percentage".into(), "nonsense-value".into())]
```
sighphyre, line 845: For the not-yet-Rust folks, this actually expects a hashmap so we make a vec of two tuples and then beat it with the `collect` stick until it becomes a hashmap

unleash-yggdrasil/src/strategy_upgrade.rs
nunogois: Should we make this method more generic and make it accept the param name as well? That way you can use for the "rollout" above as well. Perhaps it could be generic enough to also accept the type.

unleash-yggdrasil/src/strategy_upgrade.rs
nunogois: This is the least sane thing I read all week. I love it.

So if I understood correctly essentially this represents a hashmap like this:
```
{
  percentage: nonsense-value
}
```

Is this really the best way of representing this in code?

nunogois (approved):
LGTM

gastonfournier (approved):

unleash-yggdrasil/src/strategy_upgrade.rs
sighphyre: Yessir! That's spot on. This is fairly normal for Rust, collect pretty idiomatic for this.

There's two other options: Make a mut hashmap externally and populate it then pass that. But it's no simpler than this and probably less idiomatic

The other option is to use the map_lit crate for creating map literals but that adds to compile times and I'm trying to avoid it

unleash-yggdrasil/src/strategy_upgrade.rs
sighphyre: Yuh fair, I've added a param name to this so they now share code but I'm very hesitant to bind these together too tightly with an abstraction, I think this is one of those cases where the code is identical but the abstraction is a coincidence

unleash-yggdrasil/src/strategy_upgrade.rs
nunogois: TIL. Thanks for the explanation!

PR #103, opened 2023-11-29
fix(core): hostname strategy should now work correctly

sighphyre:
This fixes a few silly mistakes in the hostname strategy (my bad).

- Hostname strategy should now be generated correctly in the upgrade path, rather than creating a remote address strategy
- Hostname strategy should now completely ignore casing
- Hostname strategy in Rust land should check `hostname` and not `HOSTNAME`

### DIFF
--- a/unleash-yggdrasil/src/strategy_parsing.rs
+++ b/unleash-yggdrasil/src/strategy_parsing.rs
@@ -332,7 +332,7 @@ fn rollout_constraint(mut node: Pairs<Rule>) -> RuleFragment {
 
 fn get_hostname() -> Result<String, SdkError> {
     //This is primarily for testing purposes
-    if let Ok(hostname_env) = env::var("HOSTNAME") {
+    if let Ok(hostname_env) = env::var("hostname") {
         return Ok(hostname_env);
     }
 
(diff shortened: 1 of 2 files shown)


unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -332,7 +332,7 @@ fn rollout_constraint(mut node: Pairs<Rule>) -> RuleFragment {
 
 fn get_hostname() -> Result<String, SdkError> {
     //This is primarily for testing purposes
-    if let Ok(hostname_env) = env::var("HOSTNAME") {
+    if let Ok(hostname_env) = env::var("hostname") {
```
sighphyre, line 335: Previously HOSTNAME was n arbitrary choice but this is a more consistent way of checking this, since this is what's actually set by most OSes

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -955,23 +960,34 @@ mod tests {
...
 
-        std::env::remove_var("HOSTNAME");
+        std::env::remove_var("hostname");
+    }
+
+    #[test]
+    fn hostname_constraint_ignores_casing() {
+        std::env::set_var("hostname", "DoS");
+
+        let rule = compile_rule("hostname in [\"dOS\", \"pop-os\"]").unwrap();
```
gardleopard, line 987: is this dos with O as a number 0? i think that is confusing

gardleopard (changes requested):
test misleading to me

unleash-yggdrasil/src/strategy_parsing.rs
sighphyre: No? That's the letter "O" not the number "0".

I can change it if you think the font is misleading?

gardleopard (approved):
LGTM after that test changes

unleash-yggdrasil/src/strategy_parsing.rs
sighphyre: Swapped out for a weirdly cased `darwin`. Thanks I can never unsee this now. I'm always going to question PRs on GH that include either O or 0

unleash-yggdrasil/src/strategy_parsing.rs
gastonfournier: Should we do "hostname" or else "HOSTNAME"?

gastonfournier (approved):
LG

PR #109, opened 2023-11-30
chore: bump unleash-types and chronos

gastonfournier:
## About the changes
This addresses an issue reported in edge where strategy variants are null

https://github.com/Unleash/unleash-edge/issues/353


sighphyre (approved):
LG

If the tests are happy I'm happy

PR #110, opened 2023-11-30
feat(core): patch remote address strategy to handle subnets

sighphyre:
# What 

This adds a special constraint for remote address matching. This will check both IPv4 and and IPv6 addresses and respects subnet masks. 

# The Ugly

This builds an entirely different strategy for ip address, which treats ip addresses as strings within the parser and converts them to actual ip addresses in the compiler layer. I'd very much like to not do this, unfortunately, there's no validation in the Unleash layer to check that the ip addresses listed in the strategy are valid ip addresses, which means that treating these as ips in the parser runs the risk of breaking the entire strategy.

I think we'll have to run with this until we can deprecate the existing non validated remote address strategy

### DIFF
--- a/unleash-yggdrasil/src/strategy_parsing.rs
+++ b/unleash-yggdrasil/src/strategy_parsing.rs
@@ -3,13 +3,16 @@ extern crate pest;
 use std::collections::HashSet;
 use std::env;
 use std::io::Cursor;
+use std::net::IpAddr;
 use std::num::ParseFloatError;
+use std::str::FromStr;
 
 use crate::sendable_closures::{SendableContextResolver, SendableFragment};
 use crate::state::SdkError;
 use crate::EnrichedContext as Context;
 use chrono::{DateTime, Utc};
 use hostname;
+use ipnetwork::{IpNetwork, IpNetworkError};
 use murmur3::murmur3_32;
 use pest::error::Error;
 use pest::iterators::{Pair, Pairs};
@@ -213,6 +216,10 @@ fn string(node: Pair<Rule>) -> String {
     string.replace("\\\"", "\"")
 }
 
+fn ip(node: Pair<Rule>) -> Result<IpNetwork, IpNetworkError> {
+    IpNetwork::from_str(&string(node))
+}
+
 //Constraints
 fn numeric_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
     let context_getter = context_value(node.next().unwrap().into_inner());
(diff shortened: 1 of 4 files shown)


unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -495,6 +526,7 @@ fn constraint(mut node: Pairs<Rule>) -> RuleFragment {
         Rule::list_constraint => list_constraint(inverted, child.into_inner()),
         Rule::hostname_constraint => hostname_constraint(inverted, child.into_inner()),
         Rule::external_value => external_value(inverted, child.into_inner()),
+        Rule::ip_constraint => ip_matching_constraint(child.into_inner()),
```
sighphyre, line 529: We don't use invert for this strategy and I want to keep this simple until we can throw it away

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -990,4 +1022,50 @@ mod tests {
 
         std::env::remove_var("hostname");
     }
+
+    #[test_case("127.0.0.1", "127.0.0.1", true; "Exact match")]
```
sighphyre, line 1026: All of these are shamelessly stolen from the Java SDK test cases, which is the only SDK that deals with IPv6 and subnets

daveleek (approved):
> there's no validation in the Unleash layer to check that the ip addresses listed in the strategy are valid ip addresses, which means that treating these as ips in the parser runs the risk of breaking the entire strategy.

So the result/outcome of adding an IP that is not valid as a constraint to the strategy is that the strategy will always fail until fixed? Perhaps we should warn about this?

In general - LG. Can't say much about the code as I'm not proficient, but I guess the big benefit of Rust is that the compiler will tell you if things are wrong xD

sighphyre:
> > there's no validation in the Unleash layer to check that the ip addresses listed in the strategy are valid ip addresses, which means that treating these as ips in the parser runs the risk of breaking the entire strategy.
> 
> So the result/outcome of adding an IP that is not valid as a constraint to the strategy is that the strategy will always fail until fixed? Perhaps we should warn about this?
> 
> In general - LG. Can't say much about the code as I'm not proficient, but I guess the big benefit of Rust is that the compiler will tell you if things are wrong xD

No, the complete opposite of this. Which is why I took this approach rather than handling the ip addresses in the parser. The `harvest_ip_list` handles this by stripping things that aren't valid IP addresses before the matching is done. But this should be clearer and validated so I've added a test for this!

sighphyre:
> LGTM, but mostly looking at the tests. I'd like to test with something different than localhost, which would be more realistic, but I also think that's just details

I'm not sure I follow, there's quite a few tests for both the IPv4 and IPv6 that don't use localhost. Localhost is also just an ip address, it's not special from the perspective of what the code is doing here (at least not in Rust, I know some languages do stupid things here).

gastonfournier:
> I'm not sure I follow, there's quite a few tests for both the IPv4 and IPv6 that don't use localhost. Localhost is also just an ip address, it's not special from the perspective of what the code is doing here (at least not in Rust, I know some languages do stupid things here).

Maybe I'm just reading it wrong... or my head is out of fuel... When I read https://github.com/Unleash/yggdrasil/blob/6cf24abf4ae317f75b4a4a844319e9acadc191f5/unleash-yggdrasil/src/strategy_parsing.rs#L369-L372

I think I was assuming Ygg was reading from the local machine and then getting 127.0.0.1, but it's probably just reading from the context generated by the user in their remote machine. Right?

PR #111, opened 2023-12-04
chore(dotnet): prebuild yggdrasil on Debug build

daveleek:
## About the changes

Runs the Yggdrasil cargo build --release on dotnet build as a prebuilt step when building as debug (should help local dev)


sighphyre (approved):
Aww yiss

PR #113, opened 2023-12-05
chore(yggdrasil): add a failing test to demonstrate missing variants …

daveleek:
## About the changes

adds a failing test to demonstrate missing variants in failing strategy breaking flow

## Discussion points

I think this one is good to iron out, do we have an issue with this in other SDKs? How is this meant to work?


sighphyre (changes requested):
Very nice, thank you!

Mind applying `cargo fmt` to this before we merge?

PR #114, opened 2023-12-06
fix(core)/missing strategy variants no longer impacts other strategies

sighphyre:
Fixes an issue where strategy variants with an empty list would cause Yggdrasil to evaluate `get_variant` when it should fallback to the base toggle variants. This is done by moving the check for missing variants from rule compile time to rule execution time

Bug seems to be a regression caused by unleash-types forcing missing variants to resolve to an empty list rather than None.

### DIFF
--- a/unleash-yggdrasil/src/state.rs
+++ b/unleash-yggdrasil/src/state.rs
@@ -36,7 +36,6 @@ impl EnrichedContext {
     }
 }
 
-
 #[derive(Debug)]
 pub enum SdkError {
     StrategyEvaluationError,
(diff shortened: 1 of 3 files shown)


unleash-yggdrasil/src/state.rs
```diff
@@ -36,7 +36,6 @@ impl EnrichedContext {
     }
```
sighphyre, line 36: Everything below is formatter

daveleek (approved):
Ok, I think I grasp this. LG!

PR #117, opened 2024-01-23
chore: rework python deps to remove unneeded vulnerable packages

sighphyre:
This throws out the bulk of the Python dependencies. Most of these were a hangover from using the PyO3 project before we moved to raw FFI. These dependencies are now triggering vulnerability warnings so it's easier just to prune them down to the minimal set than attempt to update libs we don't use


andreas-unleash (approved):
LGTM

PR #118, opened 2024-01-23
chore: update types lib

sighphyre:
Done before required library update is merged, this won't build, until that's released


chriswk (approved):
Nice!

PR #119, opened 2024-02-15
fix(ruby): works on jruby

sighphyre:
This strips an unnecessary dependency from the Ruby engine that was blocking Yggdrasil from working on JRuby (pretty sure this means Rubinus will work too but we don't guarantee that as a target and I haven't and am not intending to test that). 

Most of the changes here are as a result of tooling setting up the necessary changes to pull the equivalent native interfaces here for the JVM

### DIFF
--- a/ruby-engine/Gemfile
+++ b/ruby-engine/Gemfile
@@ -5,7 +5,6 @@ source "https://rubygems.org"
 # Gemfile
 
 gem "ffi"
-gem "fiddle"
 group :development do
   gem "get_process_mem", "~> 0.2.7"
 end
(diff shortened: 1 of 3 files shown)


ruby-engine/Gemfile
```diff
@@ -5,7 +5,6 @@ source "https://rubygems.org"
 # Gemfile
 
 gem "ffi"
-gem "fiddle"
```
sighphyre, line 8: This is the problem, FFI is an alternative, this was a stale dep that should have been removed

gardleopard (approved):
4

PR #120, opened 2024-02-28
fix: isolate toggle failures

sighphyre:
This makes two patches to Yggdrasil's strategy parsing flow:

1) Toggles will no longer affect each other during compilation. Previously, a toggle with an invalid rule would fail the entire response. Now that single toggle will default to false and others will compile correctly.This should be very rare
2) The user id strategy is more tolerant to failure in it's parameters - specifically, it'll escape cases where Unleash would sometimes save and send an incorrect set of parameters

This does have the side effect of preventing major issues in the grammar from bubbling in a meaningful way. Unfortunately, standard logging frameworks do not play nicely with cross compilation to WASM and writing to std out/err can cause problems when done from guest binaries when using FFI. This is fixable, but it requires a better pattern to bubble partial failures, that's bigger than this PR.

Update: #122  Makes errors bubble correctly again

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -138,9 +138,9 @@ pub fn compile_state(state: &ClientFeatures) -> Result<HashMap<String, CompiledT
                 enabled: toggle.enabled,
                 compiled_variant_strategy: variant_rule,
                 variants: compile_variants(&toggle.variants),
-                compiled_strategy: compile_rule(rule.as_str()).map_err(|err| {
-                    SdkError::StrategyParseError(format!("Failed to parse rule, {:?}", err))
-                })?,
+                compiled_strategy: compile_rule(rule.as_str())
+                    .unwrap_or_else(|_| Box::new(|_| false)),
+
                 impression_data: toggle.impression_data.unwrap_or_default(),
                 project: toggle.project.clone().unwrap_or("default".to_string()),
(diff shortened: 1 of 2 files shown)


unleash-yggdrasil/src/lib.rs
```diff
@@ -1450,73 +1450,6 @@ mod test {
         assert_eq!(targeted_toggle.variant.name, "another");
     }
 
-    #[test]
-    pub fn invalid_date_format_bubbles_up_a_nice_error_message() {
```
sighphyre, line 1454: We no longer bubble errors so this isn't relevant

unleash-yggdrasil/src/lib.rs
```diff
@@ -1760,4 +1693,52 @@ mod test {
...
+                ]
+              },
+              {
+                "name": "This_should_be_okay",
+                "enabled": true,
+                "strategies": [
+                  {
+                    "name": "userWithId",
+                    "parameters": {
+                      "userIds": "this,is,okay"
```
daveleek, line 1722: the broken part about this is that it's an array and not a comma-separated list?

daveleek (approved):
LGTM!

unleash-yggdrasil/src/lib.rs
sighphyre: Yessir!

We expect this format: "thing,otherthing,yetanotherthing"

The array encoded format is just completely broken

PR #121, opened 2024-02-29
refactor: better handling for unwrapping grammar

sighphyre:
There's a common pattern in the rule parser where we need to unpack x number of nodes from the parent into the children nodes, e.g. the hostname constraint, which has 3 children (you can see that from the grammar itself):

`hostname_constraint = { hostname ~ in ~ string_list }`

So we see this:

```
   let name = node.next().unwrap();
   let operator = node.next().unwrap();
   let list = node.next().unwrap();
```

This is safe so long as you know the correct number of nodes and hold to that but this unwrapping pattern is error prone (you need to get exactly the correct number of nodes or risk a panic). 

This PR swaps out this pattern in the code to go through a single function that does this unwrapping, now all those unwraps area together so we can beat them with the `Result` stick.

This gets pattern gets built on in #122 to completely eliminate the panicking code

### DIFF
--- a/unleash-yggdrasil/src/strategy_parsing.rs
+++ b/unleash-yggdrasil/src/strategy_parsing.rs
@@ -46,13 +46,8 @@ pub fn normalized_hash(
     murmur3_32(&mut reader, seed).map(|hash_result| hash_result % modulus + 1)
 }
 
-fn drain<const N: usize>(mut node: Pairs<Rule>) -> [Pair<Rule>; N] {
-    let mut results = vec![];
-    for _ in 0..N {
-        results.push(node.next().unwrap());
-    }
-    let array: [Pair<Rule>; N] = results.try_into().unwrap();
-    array
+fn drain<const N: usize>(node: Pairs<Rule>) -> [Pair<Rule>; N] {
+    drain_partial(node).0
 }
 
 fn drain_partial<const N: usize>(mut node: Pairs<Rule>) -> ([Pair<Rule>; N], Pairs<Rule>) {
--- a/unleash-yggdrasil/src/strategy_parsing.rs
+++ b/unleash-yggdrasil/src/strategy_parsing.rs
@@ -46,6 +46,24 @@ pub fn normalized_hash(
     murmur3_32(&mut reader, seed).map(|hash_result| hash_result % modulus + 1)
 }
 
+fn drain<const N: usize>(mut node: Pairs<Rule>) -> [Pair<Rule>; N] {
+    let mut results = vec![];
+    for _ in 0..N {
+        results.push(node.next().unwrap());
+    }
+    let array: [Pair<Rule>; N] = results.try_into().unwrap();
+    array
+}
+
+fn drain_partial<const N: usize>(mut node: Pairs<Rule>) -> ([Pair<Rule>; N], Pairs<Rule>) {
+    let mut results = vec![];
+    for _ in 0..N {
+        results.push(node.next().unwrap());
+    }
+    let array: [Pair<Rule>; N] = results.try_into().unwrap();
+    (array, node)
+}
+
 pub type RuleFragment = Box<dyn SendableFragment + Send + Sync + 'static>;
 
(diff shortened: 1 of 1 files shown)


unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -46,6 +46,23 @@
...
+fn drain<const N: usize>(node: Pairs<Rule>) -> [Pair<Rule>; N] {
+    drain_partial(node).0
+}
+
+fn drain_partial<const N: usize>(mut node: Pairs<Rule>) -> ([Pair<Rule>; N], Pairs<Rule>) {
+    let mut results = vec![];
+    for _ in 0..N {
+        results.push(node.next().expect(&format!(
+            "Expected at least {N} elements when parsing the following pairs {node:?}"
+        )));
```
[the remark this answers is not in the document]
sighphyre: These two lints are valid, I'll fix

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -46,6 +46,23 @@ pub fn normalized_hash(
     murmur3_32(&mut reader, seed).map(|hash_result| hash_result % modulus + 1)
 }
 
+fn drain<const N: usize>(node: Pairs<Rule>) -> [Pair<Rule>; N] {
+    drain_partial(node).0
+}
+
+fn drain_partial<const N: usize>(mut node: Pairs<Rule>) -> ([Pair<Rule>; N], Pairs<Rule>) {
```
sighphyre, line 53: I don't even care if you understand this, this is just so damn cool I'm gonna talk about it.

This is generic over a compile time constant usize: the compiler will build us one implementation for every value between 0 and usize max that we use. That sounds like a lot but we only use this for 1, 2 and 3 so it's okay. This function yields a compile time checked array of size N (the constant we're generic over). 

What that allows us to do is get exactly N children from our node and enforce that at compile time.

Because Rust is awesome, the compiler will infer a ton of that information for us.

So this code:

```
let [context_getter, ordinal_operation, number] = drain(node);
```

Is just sugar for this (pseudo-code):

```
let children: [Rule; 3]= drain(node); //we remove exactly 3 nodes and unwrap them, this compile time checked
let [first, second, third] =  children;
```

So we're effectively delegating our compile time checking of the Yggdrasil DSL to the Rust compiler!

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -274,10 +298,13 @@ fn date_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
     })
 }
 
-fn semver_constraint(inverted: bool, mut node: Pairs<Rule>) -> RuleFragment {
-    let context_getter = context_value(node.next().unwrap().into_inner());
-    let ordinal_operation = to_ordinal_comparator(node.next().unwrap());
-    let semver = semver(node.next().unwrap());
+fn semver_constraint(inverted: bool, node: Pairs<Rule>) -> RuleFragment {
+    let children: [Pair<Rule>; 3] = drain(node);
```
sighphyre, line 302: Could be done in one line. I'll fix after review. Don't want to muck with the upstream branch

unleash-yggdrasil/src/strategy_parsing.rs
chriswk: Lol, `I don't even care if you understand this`. Kinda harsh? I love the result here though. :+1:

chriswk (approved):
Nice. :)

unleash-yggdrasil/src/strategy_parsing.rs
sighphyre: Haha, I'm just super excited and I want the opportunity to talk about how cool Rust is!

unleash-yggdrasil/src/strategy_parsing.rs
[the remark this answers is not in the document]
sighphyre: Actually these are closed in #122

PR #122, opened 2024-02-29
fix: improves edge case handling in rule parsing  

sighphyre:
This builds on the work in #121, to completely eliminate the unwrapping code and replaces it with an error handling pattern that allows non compilable rules to yield an error result rather than panic.

Now any node handling function that could previously panic, will now yield a Result<T, SdkError>


chriswk (approved):
LGTM

PR #123, opened 2024-02-29
chore(core): allow take_state to bubble warnings to caller

sighphyre:
Patches take_state to now return a list of warnings of toggles that failed to compile. Take state no longer indicates that it can fail because it can't

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -125,6 +125,7 @@ fn compile_variant_rule(
     variant_rules
 }
 
+#[derive(Debug, Serialize)]
 pub struct EvalWarning {
     pub toggle_name: String,
     pub message: String,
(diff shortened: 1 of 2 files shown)


unleash-yggdrasil/src/lib.rs
```diff
@@ -138,8 +148,13 @@ pub fn compile_state(state: &ClientFeatures) -> Result<HashMap<String, CompiledT
                 enabled: toggle.enabled,
                 compiled_variant_strategy: variant_rule,
                 variants: compile_variants(&toggle.variants),
-                compiled_strategy: compile_rule(rule.as_str())
-                    .unwrap_or_else(|_| Box::new(|_| false)),
+                compiled_strategy: compile_rule(rule.as_str()).unwrap_or_else(|e| {
+                    warnings.push(EvalWarning {
+                        toggle_name: toggle.name.clone(),
+                        message: format!("Failed to compile toggle, this will always be off {e:?}"),
```
gastonfournier, line 154: Maybe not off but default

gastonfournier (approved):
Looks good! I like it!

PR #128, opened 2024-06-19
chore: publish dotnet

nunogois:
https://linear.app/unleash/issue/2-1629/yggdrasilnet-package-and-publish

Adds a workflow to publish the dotnet engine.

~~**Note**: Not uploading to NuGet yet, since we need to set up a token for that.~~
Now publishing to NuGet: https://www.nuget.org/packages/Unleash.Yggdrasil

Current [package size](https://github.com/Unleash/yggdrasil/actions/runs/9615428370/job/26522725796?pr=128#step:8:8) is at:

```
-rw-r--r-- 1 runner docker 3.4M Jun 21 14:33 Yggdrasil.Engine/bin/Release/Yggdrasil.Engine.1.0.0-beta.0.nupkg
-rw-r--r-- 1 runner docker 3.4M Jun 21 14:33 Yggdrasil.Engine/bin/Release/Yggdrasil.Engine.1.0.0-beta.0.symbols.nupkg
```

Which includes:

image

Weirdly, `aarch64-unknown-linux-gnu` was the most problematic to get right.

## Manual test

**Edit**: Now that we have the package available in NuGet you can simply:

```sh
dotnet add package Unleash.Yggdrasil --version 1.0.0-beta.2
```

---

But if you're curious, here's how I manually tested this on my M1 Mac before the NuGet deployment:

1. Downloaded NuGet package artifact from the action: https://github.com/Unleash/yggdrasil/actions/runs/9615428370/job/26522725796?pr=128#step:9:21

2. Extracted the ZIP artifact and in that folder ran:

```sh
dotnet nuget push "*.nupkg" -s ~/dev/nuget # choose whatever folder you want to use as a local NuGet source
```

3. Added  NuGet locals and cleared their cache:

```sh
dotnet nuget add source ~/dev/nuget # same folder as before
```

4. Created a new dotnet project to test our engine:

```sh
dotnet new console -n dotnet-test
```

5. Installed our package in the test project:

```sh
dotnet nuget locals all --clear # just to ensure we don't use stale data
dotnet add package Yggdrasil.Engine --prerelease
```

6. Ran test project:

```cs
using System.Text.Json;
using Yggdrasil;

class Program
{
  static void Main(string[] args)
  {
    JsonSerializerOptions options = new JsonSerializerOptions
    {
      PropertyNamingPolicy = JsonNamingPolicy.CamelCase
    };

    var testDataObject = new
    {
      Version = 2,
      Features = new[] {
                new {
                    Name = "test.flag",
                    Type = "release",
                    Enabled = true,
                    ImpressionData = true,
                    Strategies = new [] {
                        new {
                            Name = "default",
                            Parameters = new Dictionary<string, string>()
                        }
                    }
                }
            }
    };

    var testData = JsonSerializer.Serialize(testDataObject, options);
    var engine = new YggdrasilEngine();
    engine.TakeState(testData);
    var featureName = "test.flag";
    var result = engine.IsEnabled(featureName, new Context());

    Console.WriteLine(result);
  }
}
```

```sh
❯ dotnet run
True
```


.github/workflows/publish-dotnet.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: What version number would you like to use? The version number should be **without** a leading `v`, e.g. `5.7.1` or `6.2.4`.
+        required: true
+
+jobs:
+  # TODO: Consider making this its own workflow to be reused by other workflows.
```
nunogois, line 12: Would be great to see this take a more modular shape, similar to the new Unleash release workflow by @gastonfournier.

I assume when we want to publish a new version of Yggdrasil we want to publish all language-specific engines, so we could have a single build action for all OS+arch we're targeting, that when finished calls all the language-specific publish actions.

However this is probably good enough for now and for the scope of this PR.

daveleek: Yeah, for a beta this should be good enough, but for production we'll need orchestration of platform specific builds along with language specific builds. but also the possibility to run separate language builds for bug fixes I imagine

daveleek (approved):
Great!

.github/workflows/publish-dotnet.yaml
sighphyre: > Would be great to see this take a more modular shape
Yes
> However this is probably good enough for now, given the scope of this PR
But also yes.

My idle long term goal was always to do as you suggest but hacky hacky hack hack is the right thing to do now while we iterate on the hard problems. Fan out builds are the the easy problem I think

dotnet-engine/README.md
```diff
@@ -1,4 +1,11 @@
-# CSharp Bindings for Yggdrasil
+# Unleash Yggdrasil .NET Engine
+
+Unleash Yggdrasil Engine for .NET.
```
sighphyre, line 3: Nice!

dotnet-engine/Yggdrasil.Engine/Yggdrasil.Engine.csproj
```diff
@@ -5,12 +5,19 @@
...
     <Version>1.0.0-beta.0</Version>
     <Company>Bricks Software AS</Company>
     <Authors>Unleash</Authors>
     <IncludeSymbols>True</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <RepositoryUrl>https://github.com/Unleash/yggdrasil</RepositoryUrl>
+    <PackageTags>unleash feature-flag feature-toggle engine</PackageTags>
+    <Description>Unleash Yggdrasil Engine for .NET. This is a .NET wrapper containing the core of the Unleash SDK domain logic, enabling you to develop your own Unleash .NET SDK.</Description>
+    <PackageIcon>worldtree_icon.png</PackageIcon>
```
sighphyre, line 18: Yeah? Cool!

sighphyre (approved):
Yeah super cool, dude!

PR #129, opened 2024-06-19
refactor: use libyggdrasilffi directly in dotnet

nunogois:
https://linear.app/unleash/issue/2-2343/use-libyggdrasilffi-directly-in-net
https://linear.app/unleash/issue/2-2344/build-net-engine-for-all-platforms

By using `libyggdrasilffi` directly in .NET we reduce the complexity we currently have when loading the library.


sighphyre (approved):
That's much simpler. That's nice. Build is green, we could use a Windows build here but I think this is worth merging

nunogois:
Finally green after some trial and error. Also noticed that Windows takes 3x more to build and test for some reason.

PR #130, opened 2024-06-26
chore: warning cleanups

sighphyre:
Silly cleanups, mostly trimming out dead code

There's one warning which we now suppress, this one is subtle but important - it allows us to identify custom strategies without treating it as an active edge case (makes the code a bit simpler to have that in place). Wouldn't be a warning if the whole stack was pure Rust but it isn't


nunogois (approved):
LGTM 👍

PR #132, opened 2024-06-26
feat: better errors in ffi layers

sighphyre:
This bubbles slightly better errors in the FFI layer so that the actual error messages are returned rather than the name of the error type. 

Also bubbles the cases where Ygg will receive a broken, uncompilable blob from an external source. Turns out we closed all the known cases where that was possible externally so... bit... hard to test now

### DIFF
--- a/dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
+++ b/dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
@@ -220,4 +220,20 @@ public class Tests
         Assert.IsTrue(result);
         Assert.IsFalse(shouldEmit);
     }
+
+    [Test]
+    public void Invalid_Json_Raises_An_Error()
+    {
+        var testData = "{\"weCloseBraces\": false";
+        var engine = new YggdrasilEngine();
+        Assert.Throws<YggdrasilEngineException>(() => engine.TakeState(testData));
+    }
+
+    [Test]
+    public void Valid_Json_With_An_Invalid_State_Update_Raises_An_Error()
+    {
+        var testData = "{\"weCloseBraces\": true}";
+        var engine = new YggdrasilEngine();
+        Assert.Throws<YggdrasilEngineException>(() => engine.TakeState(testData));
+    }
 }
--- a/yggdrasilffi/src/lib.rs
+++ b/yggdrasilffi/src/lib.rs
@@ -1,13 +1,14 @@
 use std::{
     collections::HashMap,
     ffi::{c_char, CStr, CString},
+    fmt::{self, Display, Formatter},
     str::Utf8Error,
 };
 
 use libc::c_void;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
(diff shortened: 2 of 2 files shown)


dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
```diff
@@ -220,4 +220,20 @@ public void Impression_Data_Test_Disabled()
         Assert.IsTrue(result);
         Assert.IsFalse(shouldEmit);
     }
+
+    [Test]
+    public void Invalid_Json_Raises_An_Error()
```
sighphyre, line 225: This doesn't do anything related to this PR but these tests were missing and they kinda should be here

yggdrasilffi/src/lib.rs
```diff
@@ -41,17 +42,32 @@ impl<T> From<Result<Option<T>, FFIError>> for Response<T> {
...
+    PartialUpdate(Vec<EvalWarning>),
+}
+
+impl Display for FFIError {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            FFIError::Utf8Error => write!(f, "Detected a non UTF-8 string in the input, this is a serious issue and you should report his as a bug."),
+            FFIError::NullError => write!(f, "Null error detected, this is a serious issue and you should report this as a bug."),
+            FFIError::InvalidJson(message) => write!(f, "Failed to parse JSON: {}", message),
+            FFIError::PartialUpdate(messages) => write!(
```
sighphyre, line 64: Bit hard to grok at a glance if you also haven't seen the EvalWarning code but EvalWarning is `Debug` so this becomes a list of 

{
   "toggleName": "some tooggle",
   "message": "Some error message explaining why your feature toggle is broken"  
}

In practice, this only happens if you break both the stategy_upgrade and the strategy_parsing code i.e. you need to introduce a bug that allows Ygg to accept a toggle that it can't compile.

If this does happen, it should only occur in the thread that does the toggle fetching so `isEnabled` shouldn't be affected

yggdrasilffi/src/lib.rs
```diff
@@ -41,17 +42,32 @@ impl<T> From<Result<Option<T>, FFIError>> for Response<T> {
...
-    InvalidJson,
     NullError,
+    InvalidJson(String),
+    PartialUpdate(Vec<EvalWarning>),
+}
+
+impl Display for FFIError {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            FFIError::Utf8Error => write!(f, "Detected a non UTF-8 string in the input, this is a serious issue and you should report his as a bug."),
```
nunogois, line 61:
```suggestion
            FFIError::Utf8Error => write!(f, "Detected a non UTF-8 string in the input, this is a serious issue and you should report this as a bug."),
```

nunogois (approved):

yggdrasilffi/src/lib.rs
sighphyre: :sweat_smile: Sometimes I English well but sometimes, no

Thank you

PR #134, opened 2024-07-11
fix: .NET variant constructor arg names

nunogois:
It seems like, in .NET, constructor argument names are taken into consideration when (de)serializing. This is something we stumbled upon in https://github.com/Unleash/unleash-client-dotnet/pull/224 and something that will probably bite us again later down the line.

It seems like:
 - Newtonsoft.Json wants constructor arg names to match the JSON properties;
 - System.Text.Json wants the class properties to match the constructor arg names;

This PR makes it so the Variant constructor arg names align with https://github.com/Unleash/unleash-client-dotnet/blob/main/src/Unleash/Internal/Variant.cs#L9 and then adapts the properties to match the new constructor arg names.

This change should hopefully make both serializers happy, since we support both in our .NET SDK.

This should also make it easier for us to drop `Variant.cs` in `unleash-client-dotnet` later down the line, and just rely on Yggdrasil's Variant instead, like https://github.com/Unleash/unleash-client-dotnet/pull/232

### DIFF
--- a/dotnet-engine/Yggdrasil.Engine/Types.cs
+++ b/dotnet-engine/Yggdrasil.Engine/Types.cs
@@ -33,8 +33,8 @@ public class Variant
     {
         Name = name;
         Payload = payload;
-        IsEnabled = enabled;
-        FeatureEnabled = feature_enabled;
+        Enabled = enabled;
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
LGTM

PR #135, opened 2024-07-24
feat: make go work with updated FFI

Iandenh:
## About the changes

This updates the FFI implementation of the Go code. This was outdated and was no longer working. I updated the minimum golang to 1.18 in order to use generics.

## Discussion points
- I simply use an empty object for the custom strategy (this is also now deprecated in Unleash, right?). This was also done for `getVariant` in Java
- This part (in tests) I copied from a Java test case:
```golang
 if variant == nil {
    variant = &VariantDef{"disabled", nil, false, engine.IsEnabled("Feature.A", context)}
}
```

### DIFF
--- a/go-engine/unleash_engine_test.go
+++ b/go-engine/unleash_engine_test.go
@@ -86,8 +86,9 @@ func TestGetVariant(t *testing.T) {
 	)
 
 	variant := engine.GetVariant("Feature.A", context)
+
 	if variant == nil {
-		t.Fatalf("Feature.A should have variant")
+		variant = &VariantDef{"disabled", nil, false, engine.IsEnabled("Feature.A", context)}
 	}
 	if variant.Name != "disabled" {
 		t.Fatalf("Feature.A should have been disabled")
@@ -211,6 +212,10 @@ func TestClientSpecification(t *testing.T) {
 				)
 
 				result := engine.GetVariant(toggleName, context)
+
+				if result == nil {
+					result = &VariantDef{"disabled", nil, false, false}
+				}
 
 				jsonExpectedResult, _ := json.Marshal(expectedResult)
 
--- a/go-engine/go.mod
+++ b/go-engine/go.mod
@@ -1,3 +1,3 @@
 module github.com/sighphyre/yggdrasil/go-engine
 
-go 1.17
+go 1.18
--- a/go-engine/unleash_engine.go
+++ b/go-engine/unleash_engine.go
@@ -12,17 +12,25 @@ import (
 	"unsafe"
 )
 
+const emptyCustomStrategyResults = "{}"
+
 type UnleashEngine struct {
 	ptr unsafe.Pointer
 }
 
+type response[T any] struct {
+	StatusCode   string `json:"status_code,omitempty"`
+	Value        T      `json:"value,omitempty"`
+	ErrorMessage string `json:"error_message,omitempty"`
+}
+
 func NewUnleashEngine() *UnleashEngine {
-	ptr := unsafe.Pointer(C.engine_new())
+	ptr := unsafe.Pointer(C.new_engine())
 	return &UnleashEngine{ptr: ptr}
 }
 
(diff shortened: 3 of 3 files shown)


go-engine/unleash_engine_test.go
```diff
@@ -212,6 +213,10 @@ func TestClientSpecification(t *testing.T) {
 
 				result := engine.GetVariant(toggleName, context)
 
+				if result == nil {
+					result = &VariantDef{"disabled", nil, false, engine.IsEnabled("Feature.A", context)}
```
sighphyre, line 217: I think this should just be false. Tests pass (on my machine!) with this change but if there's a reason we need to do the evaluation for the actual tests, then probably we should do something like check the actual test toggle that we're testing

```suggestion
					result = &VariantDef{"disabled", nil, false, false}
```

go-engine/unleash_engine_test.go
```diff
@@ -86,8 +86,9 @@ func TestGetVariant(t *testing.T) {
 	)
 
 	variant := engine.GetVariant("Feature.A", context)
+
 	if variant == nil {
-		t.Fatalf("Feature.A should have variant")
+		variant = &VariantDef{"disabled", nil, false, engine.IsEnabled("Feature.A", context)}
```
sighphyre, line 91: Think it's okay to have the actual toggle checked here

sighphyre (changes requested):
Oh cheers, this looks good to me. I think we should patch the test, as per my suggestion but I think this is great, thank you!

sighphyre:
Cheers, thank you so much for tackling this! Just a heads up, the FFI layers for Yggdrasil don't really run anywhere in prod at the moment. We're moving towards swapping out the code of the relevant SDKs with this in the nearish future but that's not done yet. You're more than welcome to use this if you find it useful, the core evaluation logic is very battle tested through the Edge server but the FFI layers are still a work in progress. I'd be super interested to know what you're using it for, if you are!

>     variant = &VariantDef{"disabled", nil, false, engine.IsEnabled("Feature.A", context)}
I think the Java tests might be cheating here, haha! Thanks for raising this, I'll go patch the Java side sometime in the nearish future, I've given you a suggestion for something slightly better in lieu of taking the Java version!

> I simply use an empty object for the custom strategy (this is also now deprecated in Unleash, right?).
I think this is totally fine for this PR. Custom strategies are deprecated but for us (at least on the SDK layer) that means supporting them for a long time yet. They're not that hard to add, the Ruby and C# versions have code to handle this if you're interested but I think the no-op you've put there is perfect for bringing the Go version back to a happy state for now

Iandenh:
Hi, thanks for your review and feedback. I have updated the test.

And yes, I'm using this project in a personal project. And with some modifications, it works great. Here is a bit more of a debt story about it: :smile: 

At my job, we recently moved from Overmizly to Unleash due to forcing us to migrate to another plan. When using Overmizly I was a bit unhappy with the poor web app usability and performance. To ensure I didn't need to go to that service as often, I created a small service to sit between the Overmizly CDN and my local development environment. This service can quickly toggle feature flag statuses between enabled and disabled, allowing me to test both flows efficiently.

Despite Unleash being a significant improvement in speed and usability, I refactored my application to support Unleash and act as a sort of proxy with the applied overrides (also to learn a bit about the Unleash APIs). Now, I use it as a companion tool for development. It allows me to quickly toggle feature flag statuses that only affect my environment, preventing disruptions for other developers. The tool still focuses on overriding feature flags and includes some quality-of-life features like fuzzy find and keyboard shortcuts.

I have customized this project to support a subset of the frontend API (supporting a subset of the Client API was much simpler). I found inspiration through reading Unleash Edge documentation. On my fork of this project, I added two extra FFI functions, `resolve` and `resolve_all`, which I discovered by studying the Unleash Edge frontend API implementation. If you are interested in extending the FFI functions of this library, I am more than happy to polish my code and create pull requests for them.

sighphyre (approved):
LGTM! Thank you again!

PR #137, opened 2024-08-12
fix: inverted list constraint with missing context field

### LINKED ISSUE #136: Difference unleash and yggdrasil (edge) Frontend api having an inverted constrain with missing context field
### Describe the bug

When trying stuff out, I found a difference between how Unleash handles a scenario and Yggdrasil (I also reproduced it in Unleash Edge). When you create an inverted IN check, and you don't push the context field you get feature enabled back from unleash (what I was expecting), but yggdrasil decides to put it as disabled,

### Steps to reproduce the bug

1. Create a feature flag with a strategy flexible rollout
2. Add a constrain with inverted IN with a value
3. Request with the frontend API the feature flag without sending the context field

### Expected behavior

That the feature is enabled. 

When using NOT_IN instead of an inverted IN, it works correctly.

I also checked it quickly with NUM_EQ, and there is the same problem. Unleash enables the feature yggdrasil not. And with this one, we don't have a native numeric not equals operator.

### Logs, error output, etc.

_No response_

### Screenshots

image

### Additional context

Versions:
Unleash edge build locally from this commit sha: 933659b0461b8a53cb89caf0984963918fe82b04

Same result with custom context fields and the ones already provided like appName.

### Unleash version

6.0.6

### Subscription type

None

### Hosting type

Self-hosted

### SDK information (language and version)

_No response_

Iandenh:
## About the changes
This change makes `IN` and `NOT_IN` behave the same as the Frontend API from the typescript version of Unleash without passing in the context field.

I also checked the implementation of the PHP and Go SDKs. The `IN` and `NOT_IN` constraints are always inverted there.
Some of the constraints are different. For example, in the Go SDK, the numeric and date constraints are checked to ensure that both the context and incoming constraints have the correct data format.

Here are some implementations that look like `IN` and `NOT_IN` are still inverted even when the context field is missing.
https://github.com/Unleash/unleash-client-go/blob/v4/internal/constraints/operator_in.go
https://github.com/Unleash/unleash-client-php/blob/main/src/ConstraintValidator/Operator/Lists/InListOperatorValidator.php
https://github.com/Unleash/unleash/blob/245bdeae1e044a2ef3717652d712e9249ecf70e7/src/lib/features/playground/feature-evaluator/constraint.ts#L40

Closes #136

## Discussion points
I opened this PR quickly to get a small implementation of the bug I opened here: #136. 

- This PR is to start a discussion about what is the correct implementation
- Should inverted constraints always be inverted when the context field is missing?
   - Looks like this now also differs between some SDKs

### DIFF
--- a/unleash-yggdrasil/src/strategy_parsing.rs
+++ b/unleash-yggdrasil/src/strategy_parsing.rs
@@ -468,7 +468,7 @@ fn list_constraint(inverted: bool, node: Pairs<Rule>) -> CompileResult<RuleFragm
                 match context_value {
                     Some(context_value) => {
                         let Ok(context_value) = context_value.parse::<f64>() else {
-                            return false;
+                            return false.invert(inverted);
                         };
                         match comparator {
                             ContentComparator::In => {
@@ -479,7 +479,16 @@ fn list_constraint(inverted: bool, node: Pairs<Rule>) -> CompileResult<RuleFragm
                             }
                         }
                     }
-                    None => false,
+                    None => {
+                        match comparator {
+                            ContentComparator::In => {
+                                false.invert(inverted)
+                            },
+                            ContentComparator::NotIn => {
+                                true.invert(inverted)
+                            }
+                        }
+                    },
                 }
             })
         }
@@ -491,11 +500,11 @@ fn list_constraint(inverted: bool, node: Pairs<Rule>) -> CompileResult<RuleFragm
                 match comparator {
                     ContentComparator::In => match context_value {
                         Some(context_value) => values.contains(&context_value).invert(inverted),
-                        None => false,
+                        None => false.invert(inverted),
                     },
                     ContentComparator::NotIn => match context_value {
                         Some(context_value) => !values.contains(&context_value).invert(inverted),
-                        None => true,
+                        None => true.invert(inverted),
                     },
                 }
             })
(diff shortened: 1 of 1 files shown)


sighphyre:
Thanks for the PR! I've taken a quick glance over and I think the code looks good!

What's currently uncertain to me is what the correct behavior is. There's probably two opposing arguments here:

- Inversion operators should be treated honestly regardless of their context
- Constraints that are broken should always be false (we have specification tests for this for a whole bunch of adjacent cases but not this one exactly)

I'd like to do the following before I merge this or ask for changes:

- Chat to the rest of the team about expected behavior
- Understand what all of our SDKs do in this case and related cases
- Get a client specification test down to enforce consistent behavior going forward

Just a heads up so you know you're not being ignored while I tinker!

daveleek:
Hello @Iandenh, thank you for raising this and providing us with a PR.
After a little experimentation we found that we have some inconsistencies in our SDKs around how they handle this. It's an edge case that we don't have good specifications for, so we wanted to make some decisions and standardize this across all our SDKs so they behave the same way.

Yggdrasil and its behaviour, intended as a backbone for SDKs will be "the source of truth" for our SDKs going forwards, so there may be some change in behaviour in other SDKs.

For this specific case the behaviour we decided on is that a missing context field will yield `false`, but when that constraint is inverted it will result in `true`.
We decided to move the solution a little further out so we could address all of the context field types the same way in this [PR](https://github.com/Unleash/yggdrasil/pull/139). 

We're going to close this PR and merge #139, but will keep [your issue](https://github.com/Unleash/yggdrasil/issues/136) open a little while longer.
We'll release an update as well.

Thank you for your contribution!

PR #141, opened 2024-09-05
chore: make wasm build again

nunogois:
https://linear.app/unleash/issue/2-2630/make-yggdrasil-wasm-build-again

Makes WASM build again by fixing some references and handling `hostname`, since it's not WASM-compatible.

**Question**: We're currently not running tests as part of our actions. Maybe we should?

image

image

### DIFF
--- a/unleash-yggdrasil/Cargo.toml
+++ b/unleash-yggdrasil/Cargo.toml
@@ -20,10 +20,8 @@ convert_case = "0.6.0"
 unleash-types = "0.13.0"
 chrono = "0.4.38"
 dashmap = "5.5.0"
+hostname = { version = "0.3.1", optional = true }
 ipnetwork = "0.20.0"
-
-[target.'cfg(not(target_family = "wasm"))'.dependencies]
-hostname = "0.3.1"
 
 [dependencies.serde]
 features = ["derive"]
--- a/unleash-yggdrasil/Cargo.toml
+++ b/unleash-yggdrasil/Cargo.toml
@@ -20,8 +20,10 @@ convert_case = "0.6.0"
 unleash-types = "0.13.0"
 chrono = "0.4.38"
 dashmap = "5.5.0"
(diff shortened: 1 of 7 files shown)


unleash-yggdrasil/Cargo.toml
```diff
@@ -20,9 +20,11 @@ convert_case = "0.6.0"
 unleash-types = "0.13.0"
 chrono = "0.4.38"
 dashmap = "5.5.0"
-hostname = "0.3.1"
 ipnetwork = "0.20.0"
 
+[target.'cfg(not(target_family = "wasm"))'.dependencies]
```
nunogois, line 25: This seemed more seamless than using features. WDYT?

wasm-engine/tests/web.rs
```diff
@@ -1,12 +1,12 @@
 //! Test suite for the Web and headless browsers.
 
-#![cfg(target_arch = "wasm32")]
+#![cfg(target_family = "wasm")]
```
nunogois, line 3: `target_family = "wasm"` should have more coverage than specifically `target_arch = "wasm32"`, right?

sighphyre: Sure! 64 bit WASM is still in experimental but I don't see harm in opening that up a bit

unleash-yggdrasil/Cargo.toml
sighphyre: I think it's more seamless but less explicit. I kinda want the consumer to the have to opt into turning off the feature so they know what they're losing and not have their end user discover it. Otherwise it's kinda hidden and you don't know this will get sad at you until someone runs it or you know the code intimately

sighphyre (approved):
Yuh LGTM. I still think we should move the hostname behind a flag but I think that's an unrelated detail to this PR and is safe to do later if we want

PR #142, opened 2024-09-05
chore: run wasm tests with bun

nunogois:
https://linear.app/unleash/issue/2-2634/adapt-our-yggdrasil-wasm-e2e-tests-to-use-bun-instead

Adapts our WASM e2e tests to use Bun instead. This gives us a test suite and TypeScript out of the box, and it's also faster.

I imagine we're going to add a bunch more tests here along the way, so maybe it's a worthy investment while we're still at the early stage.


ivarconr (approved):

sighphyre (approved):
Sure why not

PR #143, opened 2024-09-06
chore: implement wasm engine

nunogois:
> WIP

### DIFF
--- a/wasm-engine/Cargo.toml
+++ b/wasm-engine/Cargo.toml
@@ -15,6 +15,7 @@ default = ["console_error_panic_hook"]
 
 [dependencies]
 wasm-bindgen = "0.2"
+unleash-types = "0.13.0"
 unleash-yggdrasil = {path = "../unleash-yggdrasil"}
 getrandom = { version = "0.2", features = ["js"] }
 
--- a/wasm-engine/e2e-tests/index.test.ts
+++ b/wasm-engine/e2e-tests/index.test.ts
@@ -1,56 +1,50 @@
-import { expect, test } from 'bun:test'
-import yggdrasil from '../pkg/yggdrasil_engine'
+import { describe, beforeEach, test, expect } from 'bun:test'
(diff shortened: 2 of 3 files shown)


wasm-engine/tests/web.rs
```diff
@@ -6,45 +6,51 @@ extern crate wasm_bindgen_test;
...
-        group_id: None,
         properties: None,
-    })
-    .unwrap();
+    }).unwrap();
 
-    let result = evaluate(dsl_fragment, context).unwrap();
+    let result = engine.is_enabled("feature", context).unwrap();
     assert_eq!(result, true);
 }
```
nunogois, line 56: @sighphyre I need some insights regarding the original goal of this file. The comment at the top mentions:

> //! Test suite for the Web and headless browsers.

How so? The way I see it the only thing this test was doing was testing DSL evaluation, which I don't think belongs here.

Should we just yeet this for now?

chriswk (approved):
Looks really cool to me, but I'd wait for @sighphyre

wasm-engine/tests/web.rs
sighphyre: Ye, we can yeet it. 

These are wasm-bindgen tests, they're basically a quick and cheap way to check that your Rust code is converted to WASM correctly and executes the way you expect. The browser is only relevant insofar as its a valid WASM engine. Moving all the testing to typescript seems fine though, keeps it all in one place and removes the need for the wasm bindgen test harness which is nice

sighphyre (commented):
Yes, nice work! LGTM

sighphyre (approved):

wasm-engine/tests/web.rs
nunogois: Yep, yeeted in https://github.com/Unleash/yggdrasil/pull/143/commits/39d8bb6fd9e79d27ef0f14ca7910ef4c921a818b

PR #144, opened 2024-09-20
fix: makes random distribution better distributed

sighphyre:
This adds a way to parametize random in the grammar so that you can do things like this "random[1000]" which will create an expression of a random number bounded by 0, 1000. Default stickiness and explicitly random flexible rollouts will now use random[10000] instead of random (previously defaulting to 100), which gives random strategies better statistical distribution

### DIFF
--- a/unleash-yggdrasil/src/strategy_parsing.rs
+++ b/unleash-yggdrasil/src/strategy_parsing.rs
@@ -132,7 +132,16 @@ fn context_value(node: Pairs<Rule>) -> ContextResolver {
         Rule::remote_address => Box::new(|context: &Context| context.remote_address.clone()),
         Rule::current_time => Box::new(|context: &Context| context.current_time.clone()),
         Rule::random => {
-            Box::new(|_: &Context| Some(rand::thread_rng().gen_range(0..99).to_string()))
+            let value = child
+                .into_inner()
+                .next()
+                .map(|rule| {
+                    rule.as_str().parse::<usize>().expect(
+                        "Failed to parse random content value in the grammar, this shouldn't happen",
+                    )
+                })
+                .unwrap_or(100);
+            Box::new(move |_: &Context| Some(rand::thread_rng().gen_range(1..value).to_string()))
         }
         Rule::property => context_property(child.into_inner()),
         _ => unreachable!(),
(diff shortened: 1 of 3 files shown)


unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -18,10 +18,10 @@ use pest::Parser;
 use rand::Rng;
 use semver::Version;
 
-#[cfg(feature = "hostname")]
-use std::env;
 #[cfg(feature = "hostname")]
 use hostname;
+#[cfg(feature = "hostname")]
+use std::env;
```
thomasheartman, line 24: Seems like an arbitrary change? rustfmt update since last time?

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -371,32 +380,23 @@ fn rollout_constraint(node: Pairs<Rule>) -> CompileResult<RuleFragment> {
...
 
-        let hash = if let Some(stickiness) = stickiness {
-            normalized_hash(&group_id, &stickiness, 100, 0)
-        } else {
-            // The original code does something different here - if we're using the
-            // default strategy it generates a string of a number between 1 and 101
-            // then uses that as the hash. This instead doesn't do that and just
-            // uses a random number in place of the hash. Pretty sure it's the same thing
-            Ok(rand::thread_rng().gen_range(0..99) as u32)
-        };
```
thomasheartman, line 393: so this whole thing goes away now, yea? I take it that means that if you use random or sticky on a property that doesn't exist in the context, you get "random" stickiness instead?

thomasheartman (approved):

unleash-yggdrasil/src/strategy_parsing.rs
sighphyre: Yeah it is random, damn rustfmt chose violence this morning

unleash-yggdrasil/src/strategy_parsing.rs
sighphyre: We spoke but for anyone arriving later - turns out this code was a dead branch and this is just pruning now

PR #145, opened 2024-09-23
chore: ARM ruby build

sighphyre:
This handles `feature_enabled` for Ruby and produces a valid build for MacOS+ARM and JRuby 21 (required for ARM MacOS, which our Ruby SDK currently enforces for _reasons_)


ruby-engine/build.sh
```diff
@@ -1,4 +1,45 @@
+#!/bin/bash
```
sighphyre, line 1: You can mostly ignore this, it's a utility script so you don't have to manually copy the binaries into the Ruby folder for local dev

.github/workflows/publish-ruby.yaml
```diff
@@ -1,8 +1,5 @@
 name: Build Ruby
 on:
```
sighphyre, line 2: The build now tags the binaries produces with both their platform (via the file extension) and their arch (new, in the filename)

unleash-yggdrasil/src/lib.rs
```diff
@@ -650,6 +653,7 @@ impl VariantDef {
 }
 
 #[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Clone)]
+#[serde(rename_all = "camelCase")]
```
sighphyre, line 656: I checked Edge, this is never publicly exposed there so this should be safe

unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -430,7 +430,7 @@ mod tests {
             variants: None,
```
sighphyre, line 430: Nothing in this file is relevant, linter got an upgrade and it's decided to fix some tests

unleash-yggdrasil/src/lib.rs
```diff
@@ -529,11 +529,12 @@ impl EngineState {
         context: &Context,
         external_values: &Option<HashMap<String, bool>>,
     ) -> Option<VariantDef> {
-        self.get_toggle(name).and_then(|toggle| {
+        self.get_toggle(name).map(|toggle| {
```
sighphyre, line 532: This means that we now return the disabled variant if the feature exists but no variants are defined. This was previously not handled and was causing a spec test to fail (thank you client spec)

unleash-yggdrasil/src/lib.rs
```diff
@@ -673,11 +675,11 @@ mod test {
     use serde::Deserialize;
     use std::{collections::HashMap, fs};
     use test_case::test_case;
-    use unleash_types::client_features::{ClientFeatures, FeatureDependency, Override};
+    use unleash_types::client_features::{ClientFeatures, FeatureDependency, Override, Payload};
```
sighphyre, line 678: This and everything below in this file is just a consequence of having `feature_enabled` now properly camelCased on serialise/deserialise.

.gitignore
```diff
@@ -20,7 +20,7 @@
 /go-engine/unleash_engine.h
 /php-engine/unleash_engine.h
 /php-engine/libyggdrasilffi.so
```
gastonfournier, line 22: Should we also change this for php-engine?

ruby-engine/build.sh
gastonfournier: I would set it to fail if something goes wrong
```suggestion
#!/bin/bash
set -e
```

gastonfournier (approved):
Looks sane to me! Just some minor comments that can be dismissed

ruby-engine/build.sh
sighphyre: Thanks! Good suggestion!

.gitignore
sighphyre: Honestly I'm not sure yet. I'm don't yet understand how PHP distributes native binaries or whether or not these tricks will be needed there but definitely in a future PR

PR #146, opened 2024-09-25
chore: 0.0.5 of the ruby engine

sighphyre:
Just because doing an actual release with a beta dependency feels icky. It's time to let this thing out into the wide world


chriswk (approved):
LGTM

PR #149, opened 2024-10-22
fix: move wasm engine to handle featureEnabled not feature_enabled

sighphyre:
Only because feature_enabled is a curse and we're moving slowly towards deprecating it. Should be safe to handle in the wasm layer for now. In practice this is only patched in tests


chriswk (approved):
:1st_place_medal:

PR #151, opened 2024-10-23
chore(ruby): bump version of ffi to 1.16.3

sighphyre:


gastonfournier (approved):
Clippy is complaining but it looks good to me

PR #152, opened 2024-10-24
chore: fix feature enabled

sighphyre:
Yggdrasil core has moved on from "feature_enabled" to featureEnabled. That's not publicly exposed in Edge and the SDKs can deal with it on their own terms while we migrate forward to a consistent casing scheme.

Because this affects the JSON serialising, this affects consuming SDKs as well, in practice that's just Ruby and .NET. Ruby doesn't need a patch because it already uses snake casing conversions. This brings .NET to the table

### DIFF
--- a/dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
+++ b/dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
@@ -6,7 +6,25 @@ using System;
 using Newtonsoft.Json.Linq;
 using Yggdrasil;
 using Yggdrasil.Test;
+using System.Text.Json.Serialization;
 
+public class TestVariantReadModel
+{
+    public TestVariantReadModel(string name, Payload? payload, bool enabled, bool featureEnabled)
+    {
+        Name = name;
+        Payload = payload;
+        Enabled = enabled;
+        FeatureEnabled = featureEnabled;
+    }
+
+    public string Name { get; set; }
+    public Payload? Payload { get; set; }
+    public bool Enabled { get; set; }
+
+    [JsonPropertyName("feature_enabled")]
+    public bool FeatureEnabled { get; private set; }
+}
 
 public class Tests
 {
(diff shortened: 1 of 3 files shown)


dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
```diff
@@ -6,7 +6,25 @@
...
+        Payload = payload;
+        Enabled = enabled;
+        FeatureEnabled = featureEnabled;
+    }
+
+    public string Name { get; set; }
+    public Payload? Payload { get; set; }
+    public bool Enabled { get; set; }
+
+    [JsonPropertyName("feature_enabled")]
```
daveleek, line 25: This one seems to be missing in the Variant class?

dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -29,22 +29,20 @@ public class EngineResponse<TValue> : EngineResponse
...
+        FeatureEnabled = featureEnabled;
     }
 
     public static readonly Variant DISABLED_VARIANT = new Variant("disabled", null, false, false);
 
     public string Name { get; set; }
     public Payload? Payload { get; set; }
-    [JsonPropertyName("enabled")]
     public bool Enabled { get; set; }
-    [JsonPropertyName("feature_enabled")]
```
daveleek, line 46: Shouldn't we keep this if we add it to the test case?

daveleek (commented):
Question around a removed JsonPropertyName

dotnet-engine/Yggdrasil.Engine.Tests/YggdrasilEngineTest.cs
sighphyre: Yes, by design, the Variant class can and should handle "featureEnabled", this class only exists so we can materialise the "feature_enabled" property in the spec tests

dotnet-engine/Yggdrasil.Engine/Types.cs
sighphyre: Nah, wanna yet it. The point of this PR to stop dealing with "feature_enabled" and deal with "featureEnabled" instead so we can get consistent casing

daveleek (approved):
Fair enough, ship it! :)

PR #154, opened 2024-10-28
chore: binary upload for ffi layer

sighphyre:
Adds a binary uploader for the FFI so that we can build off known artifacts for release, rather than whatever is in the source code. Also let's us reuse the code across all engines that use the FFI layer

Probably needs WASM extensions later too but this will do for now


nunogois (approved):
LGTM!

PR #155, opened 2024-10-28
fix; make binary build actually do the upload correctly

sighphyre:
Does a few small things that makes the upload work correctly:

- Added a target for Windows ARM
- Set the URL rather than the release id in the upload artifact job
- Use different script for Windows/Unix for determining the tag
- ARM build for windows uses cross


chriswk (approved):
LGTM

PR #157, opened 2024-10-29
feat: list known features

sighphyre:
Adds a very simple way to list the features known to the engine at the time of calling. This is to support SDKs like .NET/Go/Java which have a current option to list the features and is typically used to iterate those features and evaluate them all. For the consuming code, this produces a list of objects that look like

{
  "name" "The name of the feature toggle",
  "project": "The project the feature belongs to",
  "type": "Experiment, kill switch, etc, only sent if the feature response includes this, some older formats don't"
}

Currently only implemented in the FFI layers for Ruby and .NET.

### DIFF
--- a/dotnet-engine/Yggdrasil.Engine/Types.cs
+++ b/dotnet-engine/Yggdrasil.Engine/Types.cs
@@ -72,6 +72,21 @@ public class Payload
     }
 }
 
+public class FeatureDefinition
+{
+    public FeatureDefinition(string name, string project, string? type)
+    {
+        Name = name;
+        Project = project;
+        Type = type;
+    }
+    public string Name { get; set; }
+
+    public string Project { get; set; }
+
+    public string? Type { get; set; }
+}
+
 public class YggdrasilEngineException : Exception
 {
     public YggdrasilEngineException(string message) : base(message) { }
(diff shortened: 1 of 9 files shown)


dotnet-engine/Yggdrasil.Engine/Types.cs
```diff
@@ -72,6 +72,21 @@ public override int GetHashCode()
...
+    {
+        Name = name;
+        Project = project;
+        Type = type;
+    }
+    public string Name { get; set; }
+
+    public string Project { get; set; }
+
+    public string? Type { get; set; }
```
gastonfournier, line 87: Is the type optional?

gastonfournier (approved):
LGTM! But I'm surprised the feature_type is optional

dotnet-engine/Yggdrasil.Engine/Types.cs
sighphyre: Yeeeaa... maybe. This is a good question. I've derived the type here from the types defined in our core Unleash Types rust library, which says this is optional. As for the actual question. I don't think this is optional in modern versions of Unleash but it's non existent in older versions. Not sure how far we back we have to go but there's at least one set of humans using Unleash v3 and I just haven't checked

PR #158, opened 2024-10-30
chore: update build on dotnet to use release artifacts

sighphyre:
Binary build is now handled by a completely different task that's fired on `cargo smart-release` (or just pushing a tag, but don't do that).

The .NET build now picks up the binaries from the release artifacts, and uses an internal property - YggdrasilCoreVersion in its build files to set the version of Yggdrasil FFI that will be used. This means that our versioning is now tracked in source control

### DIFF
--- a/.github/workflows/publish-dotnet.yaml
+++ b/.github/workflows/publish-dotnet.yaml
@@ -1,113 +1,45 @@
----
 name: .NET Publish
-
 on:
   workflow_dispatch:
-    inputs:
-      version:
-        description: What version number would you like to use? The version number should be **without** a leading `v`, e.g. `5.7.1` or `6.2.4`.
-        required: true
 
 jobs:
-  # TODO: Consider making this its own workflow to be reused by other workflows.
-  build-binaries:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        include:
-          - os: ubuntu-latest
-            target: x86_64-unknown-linux-gnu
-            rid: linux-x64
-            output: libyggdrasilffi.so
-          - os: ubuntu-latest
-            target: aarch64-unknown-linux-gnu
-            rid: linux-arm64
-            output: libyggdrasilffi.so
-          - os: windows-latest
-            target: x86_64-pc-windows-gnu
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-dotnet.yaml
```diff
@@ -1,135 +1,77 @@
...
-      - name: Set version
+      - name: Download and Organize Binaries
+        env:
+          RELEASE_VERSION: ${{ github.event.inputs.release_version }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          cd dotnet-engine
-          version=${{ github.event.inputs.version }}
-          sed -i "s/<Version>.*<\/Version>/<Version>$version<\/Version>/" Yggdrasil.Engine/Yggdrasil.Engine.csproj
+          binaries=("libyggdrasilffi_aarch64.so linux/aarch64/libyggdrasilffi.so"
```
sighphyre, line 32: Would love to have this list of binaries defined in one place across all builds but I can't see an obvious and sane way to do that right now

For the .NET case, we also have to put the binaries in specific locations, which makes this more challenging to abstract

Open to suggestions, but I'll probably do that in a future PR

.github/workflows/publish-dotnet.yaml
```diff
@@ -1,135 +1,77 @@
...
 
-      - name: Restore dependencies
+      - name: Extract Yggdrasil Core Version
+        id: extract_ffi_version
         run: |
-          cd dotnet-engine
-          dotnet restore dotnet-engine.sln
+          sudo apt-get install -y xmlstarlet
+          cd dotnet-engine/Yggdrasil.Engine
+          CORE_VERSION=$(xmlstarlet sel -t -v "/Project/PropertyGroup/YggdrasilCoreVersion" -n Yggdrasil.Engine.csproj)
```
sighphyre, line 23: Dirty little cheat to pull the version property for Ygg out the raw XML, because apparently msbuild doesn't give you an option to do that from the command line #sad

nunogois (approved):
Cool stuff

PR #159, opened 2024-10-30
rework ruby build to use release artifacts and disable publish

sighphyre:
Ruby build now uses published artifacts rather than assembling from scratch. We also now just overwrite all platforms when building the gem, which lets us do this on just Ubuntu rather than a bunch of different OSes

API is stabilizing now so we're moving to 0.1.0. 1.0.0 will happen when all the SDKs have been migrated

### DIFF
--- a/.github/workflows/publish-ruby.yaml
+++ b/.github/workflows/publish-ruby.yaml
@@ -3,99 +3,27 @@ on:
   workflow_dispatch:
 
 jobs:
-  build-binaries:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        include:
-          - os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-ruby.yaml
```diff
@@ -1,114 +1,61 @@
-name: Build Ruby
+name: Publish Ruby
```
nunogois, line 1: Since we're renaming these, maybe we should make them consistent across the different technologies.

nunogois (approved):
Nice!

.github/workflows/publish-ruby.yaml
sighphyre: Good suggestion, patched the .NET one so they also match their containing file

PR #162, opened 2024-11-05
fix: add musl support for .net

sighphyre:
Adds support for musl in .NET. Needed to make alpine Linux work, which is common enough to worry about


chriswk (approved):
LGTM

PR #163, opened 2024-11-05
chore: cross for more builds

sighphyre:
Swaps out the Mac and Linux cargo builds for cross. Cross more widely compatible native libraries that we link against (for the Linux case, glibc, which means we work on older Debian versions).

Apparently there's no cross target for windows x86 and so we're stuck with the cargo/cross choice rather than doing everything through cross


chriswk (approved):
As long as you manage to stay away from needing openssl, this should work fine. :+1:

PR #165, opened 2024-11-07
fix: bring python back to working state

sighphyre:
Brings the Python engine back to a working state and does some sanity clean ups

- Yggdrasil FFI function names have changed, this updates those to the new names
- Resolves the binary from ./lib, where it's expected to be for a production system + a build script to make that a bit easier to work with
- Adds typings to the public API
- Makes errors bubble
- Fixes a few leaks - we now free all the pointers correctly,

### DIFF
--- a/python-engine/build.sh
+++ b/python-engine/build.sh
@@ -1,3 +1,6 @@
 cargo build --release
 mkdir -p lib
-cp ../target/release/libyggdrasilffi.so ./lib/
+
+[ -f ../target/release/yggdrasilffi.dll ] && cp ../target/release/yggdrasilffi.dll ./lib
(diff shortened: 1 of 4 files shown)


python-engine/build.sh
```diff
@@ -0,0 +1,3 @@
+cargo build --release
+mkdir -p lib
+cp ../target/release/libyggdrasilffi.so ./lib/
```
nunogois, line 3: Should we copy all of the artifacts instead of just the `.so` one?

sighphyre: Absolutely not, people who don't use Linux deserve to suffer

...seriously though, good call, added a cp for Mac + Windows. Means git bash on Windows but that feels okayish

PR #166, opened 2024-11-07
fix: stop hostname tests from flaking

sighphyre:
Hostname tests require an env var to be set and then unset. But Rust runs tests in parallel with no isolation... yeah

This makes those flakey tests no longer flake


nunogois (approved):
I love serial in the morning!

PR #170, opened 2024-11-11
feat(python): add impression data check

sighphyre:
Adds a way for the Python engine to check whether or not a toggle has impression data enabled


chriswk (approved):
LGTM

PR #171, opened 2024-11-11
feat(python): add function to list known features

sighphyre:
Simple pass through to Ygg to list the features known to the engine at call time


chriswk (approved):
LGTM

PR #172, opened 2024-11-11
feat(python): reject registration of custom strategies that don't quack like ducks

sighphyre:
This adds a little more validation to the objects passed to the register strategy hook in Python. I've put this in place because the legacy custom strategies in the Python SDK register like this:

``` python
{
  "some-strategy": CustomStrategy
}
```

whereas, the Yggdrasil engine wants them like this:

``` python
{
  "some-strategy": CustomStrategy()
}
```
Python is spectacularly unhelpful here, and the error message you get points out that you're missing a parameter (`self`, in this case). So this moves the check from runtime to registration time so that end users will get this pointed out in testing rather than at runtime

Apparently that's a subtle enough difference that you can lose 20 minutes debugging tests so now I'm validating in anger

### DIFF
--- a/python-engine/tests/test_custom_strategy.py
+++ b/python-engine/tests/test_custom_strategy.py
@@ -1,4 +1,5 @@
 from yggdrasil_engine.custom_strategy import CustomStrategyHandler
+import pytest
 
 RAW_STATE = """
 {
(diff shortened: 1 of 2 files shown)


chriswk (approved):
Oof, I can see that being annoying. Have a :+1: to brighten your day.

PR #176, opened 2024-11-14
chore: tox build for python

sighphyre:
Creates a build and publish pipeline for Python engine. This works a little different from the other pipelines for Python reasons.

Specifically this requires some ugly shenanigans around the assembly of the wheel. This is caused by a few tools not behaving with this particular flow for binary dists. Generally the build tools around Python expect to be able to assemble the binary they require themselves and more importantly *on the architecture they support*. This is a huge pain because it requires emulation layers and docker containers. We've been through that dance already on the binary builds, which can take a lot of that pain away with cross.

So instead we do something very ugly - we allow poetry to assemble a wheel, package the binary and then we muck with the wheel until it does what it should do:
1) It advertises the correct package name {python-version}-{python-abi-version}-{platform}
2) The WHEEL metadata file in the wheel contains the correct tag and advertises that it contains a native binary

This solves two really annoying problems, first it allows package managers like pip to correctly identify the package it needs to pull for the architecture it's running on and it allows PyPI to correctly identify that the hash for the package is unique.

This unfortunately doesn't resolve the issue that we have to assemble a wheel for all the versions of Python we want to support. That's a Python detail around how binaries are loaded, and while we can support a very wide range with `abi3`, Python still wants one per version (so a total of 48 packages built and published)


python-engine/tox.ini
```diff
@@ -0,0 +1,52 @@
...
+
+    linux_arm64: PLATFORM_TAG = manylinux2014_aarch64
+    linux_arm64: BINARY = libyggdrasilffi_arm64.so
+    linux_arm64: OUTPUT_BINARY = libyggdrasilffi.so
+
+    windows_x86_64: PLATFORM_TAG = win32
+    windows_x86_64: BINARY = yggdrasilffi_x86_64.dll
+    windows_x86_64: OUTPUT_BINARY = yggdrasilffi.dll
+
+    windows_arm64: PLATFORM_TAG = win_amd64
```
nunogois, line 25: Is this correct? 🤔

nunogois (approved):
LGTM

python-engine/tox.ini
sighphyre: It's weird but yes I believe so

PR #183, opened 2024-11-26
fix: package correct binaries for windows

sighphyre:


chriswk (approved):
These seem rather important. So :+1: from me

PR #187, opened 2024-11-28
fix: memory leak in metrics

sighphyre:
Tested by just wiring it up to the SDK and evaluating a billion toggles. Leak is very obvious without this patch


daveleek (approved):
Ah good catch!

sighphyre:
> Ah good catch! What's up with the history here by the way?

I branched off the release branch (which got merged) so that I could get this out faster

PR #189, opened 2025-01-07
fix: support for utf-8 chars on win

nunogois:
https://linear.app/unleash/issue/2-3110/net-sdk-utf-8-scandinavian-character-issue

Adds support for flag names with valid UTF-8 characters like `ø` when running on Windows.

### DIFF
--- a/dotnet-engine/Yggdrasil.Engine/FFI.cs
+++ b/dotnet-engine/Yggdrasil.Engine/FFI.cs
@@ -12,19 +12,19 @@ internal static class FFI
     [DllImport("yggdrasilffi", SetLastError = true, CallingConvention = CallingConvention.Cdecl)]
     private static extern IntPtr get_metrics(IntPtr ptr);
     [DllImport("yggdrasilffi", SetLastError = true, CallingConvention = CallingConvention.Cdecl)]
-    private static extern IntPtr take_state(IntPtr ptr, string json);
+    private static extern IntPtr take_state(IntPtr ptr, byte[] json);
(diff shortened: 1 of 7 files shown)


dotnet-engine/Yggdrasil.Engine/FFI.cs
```diff
@@ -99,4 +99,16 @@ public static IntPtr ListKnownToggles(IntPtr ptr)
...
+
+    /// <summary>
+    /// Converts a string to a UTF-8 encoded null-terminated byte array.
+    /// </summary>
+    /// <param name="input">The string to convert.</param>
+    /// <returns>A UTF-8 encoded null-terminated byte array.</returns>
+    private static byte[] ToUtf8Bytes(string input)
+    {
+        byte[] utf8Bytes = System.Text.Encoding.UTF8.GetBytes(input);
+        Array.Resize(ref utf8Bytes, utf8Bytes.Length + 1);
```
sighphyre, line 111: This appears to make space for the null terminator but never gets added. Do you know what .NET does if you don't explicitly set the null byte? Can we rather just set it? It's probably fine but I see the Rust docs say that getting this wrong means UB and I'm trying to avoid that like the plague

sighphyre: Nevermind. I read the docs. This should be fine

nunogois: For more context: This should expand the existing array and append a 0 (default) byte automatically, which is what we want for our null terminator.

PR #196, opened 2025-01-09
chore(java): rework jar publish names to be compatible with maven os detector

sighphyre:
This changes the way artifact names are created for Java artifacts to be compatbile with the way [Maven OS Detector](https://github.com/trustin/os-maven-plugin) works


gastonfournier (approved):
LGTM!

PR #197, opened 2025-01-10
feat(java): expose list_known_features

sighphyre:
Exposes list_known_features in the Java layer so we can start to swap out `MoreOperations` in the Java SDK


java-engine/src/main/java/io/getunleash/engine/FeatureDef.java
```diff
@@ -0,0 +1,34 @@
...
+        this.name = name;
+        this.project = project;
+        this.type = Optional.ofNullable(featureType);
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public Optional<String> getType() {
```
chriswk, line 27: Why an optional when your constructor takes a string?

java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
```diff
@@ -89,27 +89,41 @@ void testGetVariant() throws Exception {
...
+
     @Test
     public void testClientSpec() throws Exception {
         ObjectMapper objectMapper = new ObjectMapper();
         objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
         File basePath = Paths.get("../client-specification/specifications").toFile();
         File indexFile = new File(basePath, "index.json");
-        List<String> testSuites =
-                objectMapper.readValue(indexFile, new TypeReference<List<String>>() {});
+        List<String> testSuites = objectMapper.readValue(indexFile, new TypeReference<List<String>>() {
```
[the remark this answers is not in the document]
chriswk: No worries

chriswk (approved):
Except for the Q about optional this LGTM

java-engine/src/main/java/io/getunleash/engine/FeatureDef.java
sighphyre: So I'd like the constructor to take an Optional too but apparently Jackson hates that so I figured me wanting that was wrong for some reason.

PR #198, opened 2025-01-13
refactor(java):separate resposibilities on java engine so static methods can be exposed and do memory management that should work on Java 18 and above

sighphyre:
Sorry, this one does two things because they're really difficult to separate:

Most of this is so we can do cool things like [this](https://github.com/Unleash/yggdrasil/pull/199).

### Refactor Java Engine

This splits up the responsibilities more cleanly (according to one ginger's opinions) of the Java engine. The three components do the following:

**NativeLoader**

Previously `YggdrasilFFI`, the `NativeLoader` now only holds the responsibility for loading the Yggdrasil binary into memory. This is a purely static class since there's only ever one binary and that needs to be loaded before anything else can be done. Importantly, no pointer management is done at this level. `NativeLoader` consumes the new `CamelToSnakeMapper` to provide nice camelCase names to the Java code instead of using a decorator to deal with the snake_case names.

**UnleashFFI**

This is the JNA interface to the binary. This exposes a Java interface to consume the Rust code. This does also provide a `getInstance` which returns the only valid implementation to access the binary. Consumers are free to mock this interface for testing. 

**UnleashEngine**

The code that binds the domain logic in Java to the domain logic in Rust. This _does_ do pointer management. Specifically, one UnleashEngine in Java land holds one pointer to one Yggdrasil struct in Rust land. This component therefore also takes over the tests the YggdrasilFFI was previously doing to ensure the engine was properly cleaned up 

### Dynamically Mount Cleaner Instead of Finalizer

Sooo... small problem. Finalizers are deprecated in Java 17 and marked for removal. Later JVMs provide the option to turn them off entirely, which means our existing implementation just quietly leaks. The more delightful problem is that the alternative to finalizers, the `Cleaner` API, is only Java 9 and above. And we support Java 8.

So we inject some code to check the JVM version. If you're Java 8 or below your finalizer executes as normal. If you're Java 9 or above, we use reflection to dynamically mount a cleaner and invoke that. Why reflection? Because we support source level compatibility for Java 8, which can't be compiled with direct references to `Cleaner` objects. Similar to the approach Spring uses


java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -155,4 +155,60 @@ private <T> T read(Pointer pointer, TypeReference<T> clazz) {
...
+            throw new RuntimeException("Failed to dynamically load Cleaner", e);
+        }
+    }
+
+    private void registerWithCleaner(UnleashEngine engine, Pointer enginePtr) {
+        try {
+            Class<?> cleanerClass = Class.forName("java.lang.ref.Cleaner");
+            Method registerMethod = cleanerClass.getMethod("register", Object.class, Runnable.class);
+
+            // Avoid capturing the engine itself in the lambda, otherwise this prevents GC!
```
sighphyre, line 203: I didn't spend 3 hours searching for this bug. I'm not salty at all

java-engine/src/main/java/io/getunleash/engine/CamelToSnakeMapper.java
```diff
@@ -0,0 +1,24 @@
+package io.getunleash.engine;
+
+import java.lang.reflect.Method;
+
+import com.sun.jna.FunctionMapper;
+import com.sun.jna.NativeLibrary;
+
+class CamelToSnakeMapper implements FunctionMapper {
```
chriswk, line 8: TIL about FunctionMapper

chriswk (approved):
Looks sane to me

java-engine/src/main/java/io/getunleash/engine/CamelToSnakeMapper.java
gastonfournier: +1

sighphyre: That was me yesterday haha

java-engine/src/main/java/io/getunleash/engine/CamelToSnakeMapper.java
```diff
@@ -0,0 +1,24 @@
...
+
+        for (char c : methodName.toCharArray()) {
+            if (Character.isUpperCase(c)) {
+                snakeCaseName.append('_').append(Character.toLowerCase(c));
+            } else {
+                snakeCaseName.append(c);
+            }
+        }
+
+        return snakeCaseName.toString();
```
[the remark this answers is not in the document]
sighphyre: Excellent suggestion, will do before merge

PR #199, opened 2025-01-14
feat(java): expose core version in java engine layer

sighphyre:
Exposes the Yggdrasil version in the Java layer so we can always know what binary the Java engine is shipped with. Generally useful for other engines too

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
+++ b/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
@@ -141,6 +141,11 @@ public class UnleashEngine {
         return response.getValue();
     }
 
+    public static String getCoreVersion() {
+        Pointer versionPointer = UnleashFFI.getYggdrasilCoreVersion();
+        return versionPointer.getString(0);
+    }
+
     /** Handle reading from a pointer into a String and mapping it to an object */
     private <T> T read(Pointer pointer, TypeReference<T> clazz) {
         try {
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -31,6 +31,7 @@ pub type CompiledState = HashMap<String, CompiledToggle>;
 
 pub const SUPPORTED_SPEC_VERSION: &str = "5.1.9";
 const VARIANT_NORMALIZATION_SEED: u32 = 86028157;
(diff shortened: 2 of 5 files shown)


java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -141,6 +141,11 @@ public List<FeatureDef> listKnownToggles() throws YggdrasilError {
         return response.getValue();
     }
 
+    public static String getCoreVersion() {
```
sighphyre, line 144: We're receiving a static pointer in this function so no need to clean up

unleash-yggdrasil/src/lib.rs
```diff
@@ -31,6 +31,7 @@ pub type CompiledState = HashMap<String, CompiledToggle>;
 
 pub const SUPPORTED_SPEC_VERSION: &str = "5.1.9";
 const VARIANT_NORMALIZATION_SEED: u32 = 86028157;
+pub const CORE_VERSION: &str = env!("CARGO_PKG_VERSION");
```
sighphyre, line 34: Bake the semver defined in the cargo toml file into a constant in the binary itself

yggdrasilffi/src/lib.rs
```diff
@@ -8,7 +8,12 @@ use std::{
 use libc::c_void;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use unleash_types::{client_features::ClientFeatures, client_metrics::MetricBucket};
-use unleash_yggdrasil::{Context, EngineState, EvalWarning, ExtendedVariantDef, ToggleDefinition};
+use unleash_yggdrasil::{
+    Context, EngineState, EvalWarning, ExtendedVariantDef, ToggleDefinition, CORE_VERSION,
+};
+
+static CORE_VERSION_CSTRING: std::sync::LazyLock<CString> =
```
sighphyre, line 15: I think the `expect` here is okay since this should only cause problems if this uses an escape character to inject a nul byte and this always a semver string which can't have escape characters.

For the Unrusted:

CStrings are heap allocated so we can't just make constants out of them. So we wrap this in a LazyLock, which sets the value the first time and thereafter just retrieves it. A singleton, if you will

For the Rusted:

Look ma, no `once_cell`!

chriswk (approved):
Nice

PR #200, opened 2025-01-14
fix(java): java engine encoding will always use UTF-8, even on Windows

sighphyre:
Does what it says on the tin. Does it the dumb way, because it looks like `Map.of` is Java 9 and above


gastonfournier (approved):
LGTM!

PR #204, opened 2025-01-29
feat: implement new delta api format

sjaanus:


unleash-yggdrasil/src/lib.rs
```diff
@@ -246,24 +248,14 @@ pub struct ResolvedToggle {
...
-        if warnings.is_empty() {
-            None
-        } else {
-            Some(warnings)
-        }
+    pub fn take_delta(&mut self, delta: &ClientFeaturesDelta) -> () {
+        let current_state = self.previous_state.clone();
+        let new_state = current_state.modify_and_copy(delta);
+
+        self.previous_state = new_state.clone();
```
sighphyre, line 255: Looks like `take_state` is already doing this so can we take this out?
```suggestion
```

unleash-yggdrasil/src/lib.rs
```diff
@@ -246,24 +248,14 @@ pub struct ResolvedToggle {
...
-        }
-        self.compiled_state = Some(current_state);
-        if warnings.is_empty() {
-            None
-        } else {
-            Some(warnings)
-        }
+    pub fn take_delta(&mut self, delta: &ClientFeaturesDelta) -> () {
+        let current_state = self.previous_state.clone();
+        let new_state = current_state.modify_and_copy(delta);
```
sighphyre, line 253: Since we're cloning on the previous line, can this just be modify in place now?

sighphyre (approved):
Approved so this doesn't block but won't you take a look at the comments I've left?

unleash-yggdrasil/src/lib.rs
FredrikOseberg: Ah yes. Thank you.

unleash-yggdrasil/src/lib.rs
FredrikOseberg: Looking at that more closely, it seems like I need to do a clone on `self.previous_state` unless there are other smart ways to do it I'm not aware of.

sighphyre: Something like this:

```
       let mut new_state = self.previous_state.clone();
        new_state.modify_in_place(delta);
```

PR #207, opened 2025-01-29
fix(ruby): bump yggdrasil core version to a version that actually exists

sighphyre:
Bump the Yggdrasil version referenced by Ruby so that we can publish


FredrikOseberg (approved):
LGTM

PR #209, opened 2025-01-30
chore: optimizations to core layer

sighphyre:
1) Brings the benchmark tests back to green.

2) Moves datetime fallback out of the `is_enabled` function and into the handling code for getting context values in the rule engine. This means that toggles that don't need current time don't incur the hit of resolving the current date. This improves time for basic `is_enabled` from `690.26ns` to `62.764ns` for a cool -91.017% speed improvement

### DIFF
--- a/unleash-yggdrasil/benches/benchmark.rs
+++ b/unleash-yggdrasil/benches/benchmark.rs
@@ -10,18 +10,17 @@ fn is_enabled(engine: &EngineState, toggle_name: &str, context: &Context) {
 
 fn benchmark_with_no_strategy(c: &mut Criterion) {
     let mut engine = EngineState::default();
-    engine
-        .take_state(ClientFeatures {
-            version: 2,
-            features: vec![ClientFeature {
-                name: "test".into(),
-                enabled: true,
-                ..ClientFeature::default()
-            }],
-            segments: None,
-            query: None,
-        })
-        .unwrap();
+    engine.take_state(ClientFeatures {
+        version: 2,
+        features: vec![ClientFeature {
+            name: "test".into(),
+            enabled: true,
+            ..ClientFeature::default()
+        }],
+        segments: None,
+        query: None,
+        meta: None,
+    });
     let context = Context {
         user_id: None,
         session_id: None,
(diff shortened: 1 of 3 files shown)


unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -954,7 +959,7 @@ mod tests {
 
     #[test]
     fn handles_dates_with_truncated_milliseconds() {
-        let rule = "current_time > 2022-01-29T13:00:00Z";
+        let rule = "context[\"cutoff\"] > 2022-01-29T13:00:00Z";
```
sighphyre, line 962: This test was always broken and was passing because of a short circuit. This makes the test green again

unleash-yggdrasil/src/state.rs
```diff
@@ -24,10 +24,7 @@ impl EnrichedContext {
             session_id: context.session_id.clone(),
             environment: context.environment.clone(),
             app_name: context.app_name.clone(),
-            current_time: context
```
sighphyre, line 27: This gets moved to the context resolution code, otherwise this is invoked every single time `is_enabled` is called. Turns out that's rather expensive

unleash-yggdrasil/src/state.rs
```diff
@@ -41,49 +38,3 @@ pub enum SdkError {
     StrategyEvaluationError,
     StrategyParseError(String),
 }
-
-#[cfg(test)]
```
sighphyre, line 45: Tests are consumed by the strrategy parser code now

unleash-yggdrasil/benches/benchmark.rs
```diff
@@ -10,18 +10,17 @@ fn is_enabled(engine: &EngineState, toggle_name: &str, context: &Context) {
 
 fn benchmark_with_no_strategy(c: &mut Criterion) {
     let mut engine = EngineState::default();
-    engine
```
sighphyre, line 13: Nothing interesting in this file. This fixes the benchmarks by adding the `meta` field to ClientFeatures that got missed recently and applies a `cargo fmt` to the file

nunogois (approved):
🚀

chriswk (approved):
Nice

PR #211, opened 2025-01-31
chore(java): move java publish to pack all binaries

sighphyre:
This forces the build to pack all binaries for all platforms into a single non platform specific Jar file. Turns out that maven and gradle cannot resolve transitive dependencies correctly when they're platform specific. #sad

### DIFF
--- a/java-engine/build.gradle.kts
+++ b/java-engine/build.gradle.kts
@@ -3,7 +3,7 @@ plugins {
     `maven-publish`
     signing
     id("com.diffplug.spotless") version "6.23.2"
-    id("io.github.gradle-nexus.publish-plugin") version "2.0.0"
+    id("io.github.gradle-nexus.publish-plugin").version("2.0.0")
     id("pl.allegro.tech.build.axion-release").version("1.16.0")
 }
 
--- a/java-engine/build.gradle.kts
+++ b/java-engine/build.gradle.kts
@@ -3,16 +3,13 @@ plugins {
     `maven-publish`
     signing
     id("com.diffplug.spotless") version "6.23.2"
-    id("io.github.gradle-nexus.publish-plugin").version("2.0.0")
+    id("io.github.gradle-nexus.publish-plugin") version "2.0.0"
     id("pl.allegro.tech.build.axion-release").version("1.16.0")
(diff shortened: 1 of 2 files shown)


java-engine/build.gradle.kts
```diff
@@ -3,16 +3,13 @@ plugins {
     `maven-publish`
     signing
     id("com.diffplug.spotless") version "6.23.2"
-    id("io.github.gradle-nexus.publish-plugin").version("2.0.0")
+    id("io.github.gradle-nexus.publish-plugin") version "2.0.0"
     id("pl.allegro.tech.build.axion-release").version("1.16.0")
```
daveleek, line 7: Both with and without .() appears to be in use so should be fine, but not sure what the result of it is and why this change?

sighphyre: This is the result of me desperately hacking a build file on a short time line because I want to close this project off today. They're the same line just different syntax. I don't mind changing it for consistency, feels fair... but won't you be a good citizen and suggest a change here so I can accept it? Yes I am super lazy, why do you ask?

java-engine/build.gradle.kts
```diff
@@ -3,16 +3,13 @@ plugins {
     `maven-publish`
     signing
     id("com.diffplug.spotless") version "6.23.2"
-    id("io.github.gradle-nexus.publish-plugin").version("2.0.0")
+    id("io.github.gradle-nexus.publish-plugin") version "2.0.0"
```
daveleek, line 6:
```suggestion
    id("io.github.gradle-nexus.publish-plugin").version("2.0.0")
```

java-engine/build.gradle.kts
sighphyre: Cheers! Accepted

gastonfournier (approved):
I'd suggest you enable workflow dispatch and try this on GHA before merging, but maybe you already did the test locally. Either way, let's try and find out

sighphyre:
> I'd suggest you enable workflow dispatch and try this on GHA before merging, but maybe you already did the test locally. Either way, let's try and find out
 
I'm a step ahead of you for once! Workflow dispatch is already enabled, and I've already tested it on this branch!

PR #212, opened 2025-01-31
fix(java): correctly detect musl systems and load the appropriate binaries

sighphyre:
The JVM does not list alpine correctly so we have to do a slightly more complex dance to resolve musl systems

This is precisely halfway between a dirty hack and very elegant

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/NativeLoader.java
+++ b/java-engine/src/main/java/io/getunleash/engine/NativeLoader.java
@@ -3,7 +3,10 @@ package io.getunleash.engine;
 import com.sun.jna.Library;
 import com.sun.jna.Native;
 import com.sun.jna.Pointer;
+
+import java.io.BufferedReader;
 import java.io.FileNotFoundException;
+import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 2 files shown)


java-engine/src/main/java/io/getunleash/engine/NativeLoader.java
```diff
@@ -108,6 +111,27 @@ static UnleashFFI loadLibrary() {
...
     }
 
+    // There's no reliable way ask the JVM if this is running on alpine/musl
+    // Since System.getProperty("os.name") just lists 'linux'
+    // But! Because the JVM itself is dynamically linked against either libc or
+    // musl, and we know the JVM is currently running, it must have loaded ld-musl
+    // into memory in order to run on a musl system, so we can just query the list
+    // of loaded libraries and check if ld-musl is in there
+    private static boolean isMusl() {
+        try (BufferedReader reader = new BufferedReader(new FileReader("/proc/self/maps"))) {
```
gastonfournier, line 121: `/proc/self/maps` is a file that shows the memory map of the running process that’s accessing it (“self” is basically the current process). It lists all the mapped memory regions—like shared libraries, heaps, stacks, etc.—and their permissions.

gastonfournier (approved):
LGTM! Let's try, I think it should work

chriswk (approved):
cool and hacky at the same time. :+1:

PR #213, opened 2025-02-04
feat: move state ingestion entry point

sighphyre:
Move the check for whether the inputs to take_state are a clientFeatures object or a list of delta messages.

This is done for two reasons:

1) Consumers of take_state no longer need to know or care about what message they're handling, only that it will be handled
2) Tests can be run on the public API in Yggdrasil and that's the same public API that users will use


FredrikOseberg (approved):
LGTM

chriswk (approved):
LGTM

PR #217, opened 2025-02-11
fix: support dotnet framework

sighphyre:
Add support for .NET framework.

We do this by stealing some ideas from the Java Engine - instead of using .NET core specific tools to copy the binary during the build and then load it by default we now instead dump it to the temp directory and then manually load it

### DIFF
--- a/dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
+++ b/dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
@@ -29,16 +29,75 @@ internal static class NativeLibLoader
 
     private static string GetLibraryName()
     {
-        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            return "yggdrasilffi.dll";
-        if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
-            return "libyggdrasilffi.so";
-        if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
-            return "libyggdrasilffi.dylib";
+        string os, arch, libc = "";
 
-        throw new PlatformNotSupportedException("Unsupported OS");
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            os = "win";
+        else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            os = "linux";
+        else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            os = "osx";
+        else
+            throw new PlatformNotSupportedException("Unsupported OS");
+
+        if (RuntimeInformation.ProcessArchitecture == Architecture.X64)
+            arch = "x86_64";
+        else if (RuntimeInformation.ProcessArchitecture == Architecture.Arm64)
+            arch = "arm64";
+        else if (RuntimeInformation.ProcessArchitecture == Architecture.X86)
+            arch = IntPtr.Size == 4 ? "i686" : "x86_64";
+        else
+            throw new PlatformNotSupportedException("Unsupported CPU architecture");
+
+        if (os == "linux" && IsMusl())
+            libc = "-musl";
+
+        string filename = os == "win"
+            ? $"yggdrasilffi_{arch}.dll"
+            : $"libyggdrasilffi_{arch}{libc}.{(os == "osx" ? "dylib" : "so")}";
+
+        Console.WriteLine($"[NativeLibLoader] Detected OS: {os}, Arch: {arch}, libc: {libc}");
+        Console.WriteLine($"[NativeLibLoader] Selected binary: {filename}");
+
+        return filename;
     }
 
+    private static bool IsMusl()
+    {
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            return false;
+
+        try
+        {
+            string output = File.ReadAllText("/proc/self/maps");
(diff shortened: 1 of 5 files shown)


dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
```diff
@@ -0,0 +1,179 @@
...
+            os = "osx";
+        else
+            throw new PlatformNotSupportedException("Unsupported OS");
+
+        if (RuntimeInformation.ProcessArchitecture == Architecture.X64)
+            arch = "x86_64";
+        else if (RuntimeInformation.ProcessArchitecture == Architecture.Arm64)
+            arch = "arm64";
+        else if (RuntimeInformation.ProcessArchitecture == Architecture.X86)
+            arch = IntPtr.Size == 4 ? "i686" : "x86_64";
```
sighphyre, line 48: We can't actually ask the OS if we're running on a 32 bit system because Windows will let you emulate a 32 bit system on a 64 bit arch. Soooo... we check the size of a pointer, which necessarily matches the word size

dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private static bool IsMusl()
+    {
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            return false;
+
+        try
+        {
+            string output = File.ReadAllText("/proc/self/maps");
```
sighphyre, line 69: Trick stolen from the Java SDK, if the CLR runs on a Linux system and that system is a MUSL system then we must have loaded the MUSL binaries to load the CLR, so the process space must contain something for MUSL here

dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
```diff
@@ -0,0 +1,179 @@
...
+                }
+            }
+            catch
+            {
+                return false;
+            }
+        }
+    }
+
+    internal static IntPtr LoadFunctionPointer(IntPtr libHandle, string functionName)
```
sighphyre, line 99: This method and the ones below start by trying to tap into the binary in the sane .NET core method and if we fail then we're probably running .NET framework and we fallback to kernel32.dll or libdl

dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
```diff
@@ -0,0 +1,179 @@
...
+internal static class NativeLibLoader
+{
+    internal static IntPtr LoadNativeLibrary()
+    {
+        var libName = GetBinaryName();
+        var tempPath = Path.Combine(Path.GetTempPath(), libName);
+        var assembly = Assembly.GetExecutingAssembly();
+        var assemblyName = assembly.GetName().Name;
+
+        if (!File.Exists(tempPath))
```
sighphyre, line 13: Stolen from Java - we extract the binary to a temp folder and then use built in tools to mount it. Gotta do this because .NET framework won't respect binaries in the nuget package itself

dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
daveleek: Ok, so `Architecture.X64` and `Architecture.X86` is the same thing when pointer size is 8?

dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            var loadMethod = nativeLibraryType.GetMethod("Load", new[] { typeof(string) });
+            if (loadMethod != null)
+                return (IntPtr)loadMethod.Invoke(null, new object[] { libPath });
+        }
+
+        return LoadLibrary(libPath);
+    }
+
+#if NETSTANDARD
```
daveleek, line 158: Would any of this be easier if we used compiler directives and went back to outputting .NET Framework targets? (ie `net472` et al)

dotnet-engine/Yggdrasil.Engine/Yggdrasil.Engine.csproj
```diff
@@ -6,8 +6,8 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <PackageId>Unleash.Yggdrasil</PackageId>
-    <Version>1.0.4</Version>
-    <YggdrasilCoreVersion>0.14.2</YggdrasilCoreVersion>
+    <Version>1.0.5-beta.1</Version>
```
daveleek: Are we bumping this when merging? appears it's already been sent up to NuGet.. 🤷

daveleek: someone's not paying attention - disregard!

dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
sighphyre: Yep

dotnet-engine/Yggdrasil.Engine/NativeLoader.cs
daveleek: Fair enough! Was worth asking the question though in case it could help reduce complexity

PR #218, opened 2025-02-14
perf: move .NET to use low level buffers instead of the JSON API

sighphyre:


yggdrasilffi/src/json_api.rs
```diff
@@ -0,0 +1,618 @@
+use std::{
```
sighphyre, line 1: There are no real changes here. This is breaking out the contents of "lib" into its own space. No constants/functions were harmed in the making of this file

PR #219, opened 2025-02-18
fix: metric bucket start time should be reset everytime it is sent

ivarconr:
A potential fix for the "start" property of the metrics bucket not being reset correctly. 

1. This change lacks tests, HELP WANTED! 
2. What is the SDK supposed to do if it is unable to send the metrics right now? (Might be solved outside of this fix)


sighphyre (approved):
Nice catch, I've added a test to this

Good call on the missed metrics too. The SDK itself should hang onto that object and buffer it

PR #225, opened 2025-03-05
chore: new publish WASM workflow attempt

nunogois:
https://linear.app/unleash/issue/2-3335/create-a-publish-workflow-for-our-wasm-package

New attempt for our publish WASM package workflow.

This one uses steps from https://github.com/Unleash/.github/blob/main/.github/workflows/npm-release.yml instead of trying to use it directly. I believe the reason why it was [failing](https://github.com/Unleash/yggdrasil/actions/runs/13676066854/job/38236797932) was because we were generating the build artifacts in this job and then assuming the next one would be able to pick them up. Unless we upload the artifacts and adapt our template workflow to download them, which I would prefer not doing, that probably won't ever work. 

This new approach may be more straightforward and give us more control anyways.

### DIFF
--- a/.github/workflows/publish-wasm.yaml
+++ b/.github/workflows/publish-wasm.yaml
@@ -14,10 +14,35 @@ on:
         default: 'latest'
 
 jobs:
-  build-wasm:
+  build-and-publish-wasm:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


sighphyre (approved):
LGTM

PR #226, opened 2025-03-06
chore: javascript engine

nunogois:
https://linear.app/unleash/issue/2-2641/create-a-ts-wrapper-around-our-wasm-layer

Implements our JavaScript engine, which is a simple wrapper around our WASM package for ease-of-use and type declarations.

I don't expect this to be anywhere near "finished". But it seems like a good basic first step to start experimenting before trying to shape it into something more suitable to our needs. I expect some back and forth between this and the Node SDK if we want to introduce this engine without many breaking changes.

I used Bun because it's pretty fast and the easiest way to get sane modern defaults on a new TS project IMO. I'm happy to fight myself here if I face any issues down the line. But I'm also happy if we decide that it's better to go with something else.


javascript-engine/src/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      context
+    )
+
+    const response = this.engine.checkEnabled(
+      toggleName,
+      context,
+      customStrategyPayload
+    )
+
+    return readResponse<boolean>(response) || false
```
sighphyre, line 57: This might cause you problems when you start bolting this into the NodeSDK. The SDKs tend to need to know if a toggle is missing during evaluation because they use if to doing fallback values

javascript-engine/src/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      context
+    )
+
+    const response = this.engine.checkVariant(
+      toggleName,
+      context,
+      customStrategyPayload
+    )
+
+    return (
```
sighphyre, line 72: Same thing here. The SDKs will let you set a fallback variant if the variant is completely missing, I don't think doing a fixed disabled will work here

javascript-engine/src/utils.ts
```diff
@@ -0,0 +1,24 @@
+type EngineResponse = {
```
sighphyre, line 1: Personal feeling but "utils" is where good project structure goes to die. This file looks like it has a pretty focused intent. Can we name it for what it does? I want to say this is doing WASM interop so.... what about wasm-interop?

javascript-engine/src/strategies.ts
```diff
@@ -0,0 +1,95 @@
+import type { Context, State, StrategyDefinition } from './types'
```
sighphyre, line 1: Can I ask for some tests here? This is likely going to break in the future when this needs to handle deltas and it would be super cool if a random person got warned of that

yggdrasilwasm/src/lib.rs
```diff
@@ -8,6 +8,17 @@ use wasm_bindgen::prelude::*;
 
 type CustomStrategyResults = HashMap<String, bool>;
 
+const KNOWN_STRATEGIES: [&str; 8] = [
```
sighphyre, line 11: This looks like a copy of the YggdrasilFFI version. Can we just move both of these declarations into Yggdrasil itself and then they can both just reference that?

javascript-engine/src/index.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    this.strategies = new Strategies(knownStrategies)
+
+    if (strategies != null) {
+      this.strategies.registerCustomStrategies(strategies)
+    }
+  }
+
+  public takeState(state: State): void {
+    const response = this.engine.takeState(state)
```
sighphyre, line 39: I believe this beast will return you some warnings if the user sent it state that it didn't parse correctly. I think we should attempt to bubble that up or log it

sighphyre (approved):
Yeah cool, this is pre 0.1.0 so I think its safe enough to experiment with

javascript-engine/src/index.ts
nunogois: Makes sense, better to address it now. https://github.com/Unleash/yggdrasil/pull/226/commits/dab0c08de5fc5f9e841028e4fdf63350c76da52d

javascript-engine/src/utils.ts
nunogois: Are you saying I should I build an util in utils to better organize our utils by util category?

Sorry.

I 100% agree but naming is hard. I like your suggestion 👍 https://github.com/Unleash/yggdrasil/pull/226/commits/bc84e4aff483fbb39a4de9ec0047d8567d0c0fbb

yggdrasilwasm/src/lib.rs
nunogois: Yep, good point! I assume something like this will just work: https://github.com/Unleash/yggdrasil/pull/226/commits/fceb36e422a89b3ecf3c9815430b6bdd543fd9e3

javascript-engine/src/index.ts
sighphyre: Awesome, cheers!

javascript-engine/src/index.ts
sighphyre: Awesome, cheers!

javascript-engine/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+import type {
+  Context,
+  FeatureDefinition,
+  MetricsBucket,
+  State,
+  Variant,
+  Strategy,
+  DeltaState
+} from './types'
+export * from './types'
```
nunogois, line 13: I don't think these types are sane enough for external usage. They are mostly internal and incomplete.

 Should we leave them out for now until we figure out the exact types we want to expose?

```suggestion
```

sighphyre: You'll definitely need some of them but I think that's sane. I support having a public API that's as narrow as possible!

sighphyre (approved):
Nice patch to the strategies handling code

javascript-engine/src/strategies.ts
nunogois: Addressed the delta handling here: https://github.com/Unleash/yggdrasil/pull/226/commits/df93d7dc4911d570ef04ec299c1e2d298b3e55f4

PR #227, opened 2025-03-10
Fix/musl assemblies missing

daveleek:
we missed the .so part of the musl binaries


sighphyre (approved):
Oooo... nice catch, thank you

PR #230, opened 2025-03-11
feat(java): expose enabled on list features in core and java

sighphyre:
Exposes the environment `enabled` property for a toggle on list features. Folks on the Java SDK are asking for it back


java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
```diff
@@ -260,6 +261,7 @@ void customStrategiesRequired_whenNotConfigured_returnsFalse(
         assertEquals(expectedIsEnabled, result);
     }
 
+    @SuppressWarnings("unused")
```
sighphyre, line 264: Not relevant to this PR but this warning was annoying me. The warning is raised because there's an `unused` PhantomReference in this test, which is absolutely important for the test to run

gastonfournier (approved):
LGTM!

PR #231, opened 2025-03-19
Java: Custom strategies resolved in getVariant

mturlo:
## About the changes

Hello!

My main use case for using Unleash with the Java Client is `evaluateAllToggles()`. After migrating to the 10.x version, I was puzzled about why some of my tests kept failing. After some digging, I think I've pinpointed the issue: custom strategies aren't resolved on `getVariant()` calls, which is what `evaluateAllToggles()` uses underneath. This PR contains a fix + a test case that covers this. 

Also added a build fix for ARM-based Macs as I've had some issues running the test suite initially ("aarch64" vs "arm64" returned as architecture leading to incorrectly named native package file).

## Discussion points

I've had a brief look at other client implementations in the repo (Ruby, C# etc.) and I _think_ the intention is to resolve custom strategies on `getVariant` as well, but that is only my assumption - please feel free to correct me. I assume the `EMPTY_STRATEGY_RESULTS` is an unintentional leftover from an earlier stage of development. Having the custom strategies resolved is also the behaviour present in the 9.x branch of the Unleash Java Client.

### DIFF
--- a/java-engine/build.gradle.kts
+++ b/java-engine/build.gradle.kts
@@ -62,7 +62,7 @@ val copyTestBinary = tasks.register<Copy>("copyTestBinary") {
     val targetPath = file("build/resources/test/native")
 
     val binaryName = when {
-        os.contains("mac") && platform.contains("arm") -> "libyggdrasilffi_arm64.dylib"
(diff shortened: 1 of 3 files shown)


sighphyre (approved):
Thank you! Yes this looks good, appreciate the patch here

PR #234, opened 2025-03-24
fix(java): dont mount null object custom strategies for built in stategies

sighphyre:
Stops the custom strategies handler from mounting always-false custom strategies for built in strategies. In practice this isn't a problem because the Rust engine layer will simply ignore them but it does lead to some rather scary log warnings that the strategies will be disabled

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
+++ b/java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
@@ -19,17 +19,23 @@ class CustomStrategiesEvaluator {
   private static final Logger log = LoggerFactory.getLogger(CustomStrategiesEvaluator.class);
   static final String EMPTY_STRATEGY_RESULTS = "{}";
   private final Map<String, IStrategy> registeredStrategies;
+  private final Set<String> builtinStrategies;
 
   private final IStrategy fallbackStrategy;
   private final ObjectMapper mapper;
 
   private Map<String, List<MappedStrategy>> featureStrategies = new HashMap<>();
 
-  public CustomStrategiesEvaluator(Stream<IStrategy> customStrategies) {
-    this(customStrategies, null);
+  public CustomStrategiesEvaluator(
+      Stream<IStrategy> customStrategies, Set<String> builtinStrategies) {
+    this(customStrategies, null, builtinStrategies);
   }
 
-  public CustomStrategiesEvaluator(Stream<IStrategy> customStrategies, IStrategy fallbackStrategy) {
+  public CustomStrategiesEvaluator(
+      Stream<IStrategy> customStrategies,
+      IStrategy fallbackStrategy,
+      Set<String> builtinStrategies) {
+    this.builtinStrategies = builtinStrategies;
     this.mapper = new ObjectMapper();
     this.mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
     this.registeredStrategies =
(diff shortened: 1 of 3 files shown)


java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
```diff
@@ -63,13 +63,15 @@ public void loadStrategiesFor(String toggles) {
...
-    private List<MappedStrategy> getFeatureStrategies(FeatureDefinition feature) {
+    List<MappedStrategy> getFeatureStrategies(FeatureDefinition feature) {
         List<MappedStrategy> mappedStrategies = new ArrayList<>();
         int index = 1;
         for (StrategyDefinition strategyDefinition : feature.strategies) {
-            IStrategy impl =
-                    Optional.ofNullable(registeredStrategies.get(strategyDefinition.name))
-                            .orElseGet(() -> alwaysFalseStrategy(strategyDefinition.name));
+            if (builtinStrategies.contains(strategyDefinition.name)) {
+                continue;
```
sighphyre, line 71: This check is actually the salient point here. If we find our strategy matches a built in strategy, just continue

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -63,9 +63,10 @@ public UnleashEngine(List<IStrategy> customStrategies, IStrategy fallbackStrateg
                                         }
                                         return true;
                                     }),
-                    fallbackStrategy);
+                    fallbackStrategy, new HashSet<String>(builtInStrategies));
         } else {
-            this.customStrategiesEvaluator = new CustomStrategiesEvaluator(Stream.empty(), fallbackStrategy);
+            this.customStrategiesEvaluator = new CustomStrategiesEvaluator(Stream.empty(), fallbackStrategy,
+                    new HashSet<String>());
```
sighphyre, line 69: This code path doesn't really matter for custom strategies since they'll be ignored

java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
```diff
@@ -61,10 +67,13 @@ public void loadStrategiesFor(String toggles) {
     }
   }
 
-  private List<MappedStrategy> getFeatureStrategies(FeatureDefinition feature) {
+  List<MappedStrategy> getFeatureStrategies(FeatureDefinition feature) {
     List<MappedStrategy> mappedStrategies = new ArrayList<>();
     int index = 1;
     for (StrategyDefinition strategyDefinition : feature.strategies) {
+      if (builtinStrategies.contains(strategyDefinition.name)) {
```
sighphyre, line 74: This check is actually the salient point here. If we find our strategy matches a built in strategy, just continue

java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
```diff
@@ -125,25 +134,25 @@ private VersionedFeatures(@JsonProperty("features") List<FeatureDefinition> feat
     }
   }
 
-  private static class FeatureDefinition {
+  static class FeatureDefinition {
```
sighphyre, line 137: I've exposed a few classes and methods as default access modifier so that we can poke them a little in tests. It isn't ideal to expose this kind of thing for testing purposes but a) it's not public b) making this testable without exposing these is a bigger more fragile change

gastonfournier (approved):
LG!

PR #237, opened 2025-04-23
feat: complete function for checking if a toggle is enabled

sighphyre:
Does a few small clean up things:

- Completes the context fields so we have matching values for what Unleash expects
- Moves the flatbuffer messaging into a place where the namespace matches without fiddling
- Adds counting for metrics


java-engine/src/main/java/io/getunleash/engine/WasmEngine.java
```diff
@@ -58,16 +85,10 @@ public static byte[] buildContext(String toggleName, Map<String, String> propert
                 .mapToInt(Integer::intValue)
                 .toArray();
 
-        int userIdOffset = builder.createString("some-user-called-greg");
```
daveleek, line 61: Greg got axed

daveleek (approved):
Nice one!

PR #239, opened 2025-04-24
chore: fix wasm-related java-engine build failure

daveleek:
Attempts to fix the java-engine build errors related to pure-was not being compiled yet


.github/workflows/build-java.yaml
```diff
@@ -36,6 +36,9 @@ jobs:
         uses: Swatinem/rust-cache@v2
       - name: Build yggdrasil
```
sighphyre, line 37: Could we take this one out? We don't need this anymore and it just adds to the build time

sighphyre (approved):
Cheers

PR #240, opened 2025-04-24
fix: allow runtime hostname to be specified in context

sighphyre:
This reworks some of the public APIs to take the EnrichedContext directly during flag evaluation. We do this so that we can sneak in a new parameter - runtime_hostname on the enriched context. That in turn we use for the hostname strategy in cases where we don't have access to IO - e.g. WASM.

### DIFF
--- a/pure-wasm/src/lib.rs
+++ b/pure-wasm/src/lib.rs
@@ -6,16 +6,18 @@ use std::{
     mem, slice,
 };
 
+#[allow(clippy::all)]
 mod messaging {
     #![allow(dead_code)]
     #![allow(non_snake_case)]
+    #![allow(warnings)]
     include!("enabled-message_generated.rs");
 }
 
 use flatbuffers::FlatBufferBuilder;
 use messaging::messaging::ResponseBuilder;
 
-use unleash_yggdrasil::{Context as YggContext, EngineState};
+use unleash_yggdrasil::{Context as YggContext, EngineState, state::EnrichedContext};
 
 use getrandom::register_custom_getrandom;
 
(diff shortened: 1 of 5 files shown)


pure-wasm/src/lib.rs
```diff
@@ -157,7 +159,8 @@ pub extern "C" fn check_enabled(engine_ptr: i32, message_ptr: i32, message_len:
         };
 
         let engine = &mut *(engine_ptr as *mut EngineState);
-        let enabled = engine.check_enabled(toggle_name, &context, &None);
+        let enriched_context = EnrichedContext::from(context, toggle_name.to_string(), None);
```
sighphyre, line 162: Coming in a future PR - we can now set runtime_hostname via the Java code

unleash-yggdrasil/src/lib.rs
```diff
@@ -404,9 +412,9 @@ impl EngineState {
                 return false;
             }
 
-            let parent_enabled = self.enabled(compiled_parent, context, &None); //parent toggles explicitly don't support custom strategies
+            let parent_enabled = self.enabled(compiled_parent, &context);
```
sighphyre, line 415: We now explicitly allow custom strategies for parent toggles. I think this is probably okay, but feel free to sanity check me here

unleash-yggdrasil/src/state.rs
```diff
@@ -29,6 +31,7 @@ impl EnrichedContext {
             properties: context.properties,
             external_results,
             toggle_name,
+            runtime_hostname: None,
```
sighphyre, line 34: The magic - generally if we're using the old paths, we don't need this property

This could benefit from some refactor in a future PR but this is big enough already

unleash-yggdrasil/src/strategy_parsing.rs
```diff
@@ -418,7 +426,10 @@ fn get_hostname() -> CompileResult<String> {
 }
 
 #[cfg(not(feature = "hostname"))]
-fn get_hostname() -> CompileResult<String> {
+fn get_hostname(context: &Context) -> CompileResult<String> {
+    if let Some(hostname) = &context.runtime_hostname {
```
sighphyre, line 430: The other piece of magic. If we have a hostname specified on the enriched context, we should use it

yggdrasilffi/src/lib.rs
```diff
@@ -10,8 +10,8 @@ use libc::c_void;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use unleash_types::client_metrics::MetricBucket;
```
sighphyre, line 11: Just updating the FFI layer to use the new API

daveleek (approved):
LG

PR #241, opened 2025-04-24
feat: implement pure-wasm get metrics and proper bucket serialization

daveleek:
implements serialization of metric buckets into the pure-warm library, and a java end to end test for it.
also implements variant stats for features, but since getVariant isn't implemented yet it will need to be tested and amended after that is in place


pure-wasm/src/lib.rs
```diff
@@ -201,7 +225,7 @@ pub extern "C" fn get_metrics(engine_ptr: i32, close_time: i64) -> u64 {
     unsafe {
         let engine = &mut *(engine_ptr as *mut EngineState);
         let metrics = engine.get_metrics(DateTime::from_timestamp_nanos(close_time));
-        let response = build_metrics_response(metrics);
+        let response = build_metrics_response(metrics.unwrap());
```
sighphyre, line 228: Pretty sure you don't want to crash if there's no metrics recorded yet!

pure-wasm/src/lib.rs
```diff
@@ -137,17 +137,41 @@ pub fn build_response(enabled: Option<bool>, error: Option<&str>) -> Vec<u8> {
...
+                toggle_entry_builder.add_value(toggle_value);
+                toggle_entry_builder.add_key(toggle_key);
+                toggle_entry_builder.finish()
+            }).collect();
+            let toggle_vector = builder.create_vector(&items);
             let mut resp_builder = MetricsBucketBuilder::new(&mut builder);
-            if let Some(metrics) = metrics {
-                resp_builder.add_count(metrics.toggles.iter().count() as i32);
-            }
+                resp_builder.add_start(metrics.start.timestamp_nanos_opt().unwrap());
```
sighphyre, line 172: This will crash on 2262-04-11T23:47:16.854775807

...I think that's okay xD

sighphyre (approved):
Very nice!

PR #242, opened 2025-04-24
fix: make is enabled work

sighphyre:
Makes isEnabled work correctly against the spec tests


java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -69,26 +87,41 @@ public static byte[] buildMessage(String toggleName, Context context) {
         context.getEnvironment() != null ? builder.createString(context.getEnvironment()) : 0;
 
     List<Map.Entry<String, String>> entries = new ArrayList<>(context.properties.entrySet());
-    int[] propertyOffsets = new int[entries.size()];
-    for (int i = 0; i < entries.size(); i++) {
-      Map.Entry<String, String> entry = entries.get(i);
+    List<Integer> offsets = new ArrayList<>();
+    for (Map.Entry<String, String> entry : entries) {
```
sighphyre, line 91: Small adjustment to the context properties code here - basically strip out keys with null values. We have a spec test that passes this context:
```
            "context": {
                "properties": {
                    "customFieldMissing": null
                }
            }
```

pure-wasm/src/lib.rs
```diff
@@ -201,7 +202,10 @@ pub extern "C" fn check_enabled(engine_ptr: i32, message_ptr: i32, message_len:
 
         let toggle_name = ctx.toggle_name().expect("You need to pass a toggle name and you also need to remove this expect before production!");
 
-        let context = YggContext {
+        let context = EnrichedContext {
```
sighphyre, line 205: Make use of the work in the last PR - we can now build a context directly and populate it with runtime_hostname

pure-wasm/src/logging.rs
```diff
@@ -0,0 +1,52 @@
+use core::cell::UnsafeCell;
```
sighphyre, line 1: This does some things. This is probably "trust-me-bro" code. But long story short. We allocate a 4096 byte buffer at startup. Into that buffer we can write whatever we feel like from the Rust side. On the Java side, we can read from that buffer whenever we feel like it. This gives us a way to log across the boundary and a path towards panic hooks that aren't crashy.

The `wasm_log` marco rule is just some sugar so you can write

`wasm_log!("Some message here, which emits the contents of this type: {:?}", some_variable)`

daveleek (approved):
Cool!

PR #243, opened 2025-04-25
chore: fix wrong usings of epoch millis/nanos

daveleek:
Fixes the issue of passing seconds to rust which then assumes it's nanos, sends nanos back, which we then parse in as milliseconds... now it's milliseconds all the way!


sighphyre (approved):
LGTM

PR #244, opened 2025-04-25
fix: make variants work

sighphyre:
Relatively straight forward. No surprises here I think. Definitely ignore the flatbuffer genned files for review reasons


java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
```diff
@@ -157,33 +157,32 @@ public void testClientSpec() throws Exception {
...
-
-      //     assertEquals(
-      //         expectedResultJson,
-      //         resultJson,
-      //         String.format(
-      //             "[%s] Failed test '%s': expected %b, got %b",
-      //             suiteData.name, test.get("description"), expectedResultJson, resultJson));
-      //   }
-      // }
+      List<Map<String, Object>> variantTests = suiteData.variantTests;
```
sighphyre, line 160: Restore the variant spec tests

pure-wasm/src/lib.rs
```diff
@@ -235,6 +273,62 @@ pub extern "C" fn check_enabled(engine_ptr: i32, message_ptr: i32, message_len:
     }
 }
 
+#[unsafe(no_mangle)]
+pub extern "C" fn check_variant(engine_ptr: i32, message_ptr: i32, message_len: i32) -> u64 {
```
sighphyre, line 277: I think there's a harvested API making itself clear in this file but this is probably okay for this PR

daveleek (approved):
Yup!

PR #245, opened 2025-05-05
chore: refactor resolving pointers to objects on the java side

sighphyre:
Makes the pointer code a little easier to grok and a little harder to break by wrapping it in its own functions

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
+++ b/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
@@ -21,6 +21,7 @@ import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 import messaging.ContextMessage;
 import messaging.MetricsBucket;
 import messaging.PropertyEntry;
@@ -64,7 +65,7 @@ public class UnleashEngine {
     return hostname;
   }
 
-  public static byte[] buildMessage(String toggleName, Context context) {
+  private static byte[] buildMessage(String toggleName, Context context) {
     FlatBufferBuilder builder = new FlatBufferBuilder(1024);
 
     int toggleNameOffset = builder.createString(toggleName);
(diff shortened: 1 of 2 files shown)


java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
```diff
@@ -62,30 +62,29 @@ void testIsEnabled() throws Exception {
     assertTrue(result);
   }
 
-  // @Test
-  // void testIsEnabledWithoutValidResponse() throws Exception {
-  // engine.takeState(simpleFeatures);
+  @Test
```
sighphyre, line 65: Not actually relevant to this PR but they are green now so I'm re-enabling them

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -251,50 +227,83 @@ public VariantDef getVariant(String toggleName, Context context)
...
+    T response = decoder.apply(buf);
     deallocResponseBuffer.apply(ptr, len);
-    if (bucket.togglesVector() == null) {
-      return null;
-    }
-    return bucket;
+    return response;
+  }
+
+  private <T> T callWasmFunctionWithResponse(
```
sighphyre, line 290: The important part - now this is the primary way to call into the WASM Ygg layer rather then rewriting it each time

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -251,50 +227,83 @@ public VariantDef getVariant(String toggleName, Context context)
...
-    deallocResponseBuffer.apply(ptr, len);
-
-    return new VariantDef(variant.name(), payload, variant.enabled(), variant.featureEnabled());
   }
 
   public MetricsBucket getMetrics() throws JsonMappingException, JsonProcessingException {
     ZonedDateTime now = ZonedDateTime.now(ZoneOffset.UTC);
+
     long packed = (long) getMetrics.apply(this.enginePointer, now.toInstant().toEpochMilli())[0];
+    MetricsBucket bucket = derefWasmPointer(packed, MetricsBucket::getRootAsMetricsBucket);
```
daveleek, line 255: Generics inferred from the provided Function?

daveleek (approved):
Lovely!

PR #247, opened 2025-05-05
refactor: trait for serialization

sighphyre:
Refactors the Rust -> Java serialization layer to use a trait for serialization. This has only one important reason for existing - the boiler plate for accessing the thread local builder, clearing the buffer and materializing the byte stream are now difficult to forget to do. That's pretty important since forgetting those things means a corrupted byte stream.

As a nice side effect, it makes the top level API much nicer, since you can do things like this

`let byte_response = SomeFlatBufferType::build_response(some_rust_struct)`

And the compiler will figure the correct dispatch thanks to monomorphization

### DIFF
--- a/pure-wasm/src/lib.rs
+++ b/pure-wasm/src/lib.rs
@@ -6,6 +6,7 @@ use std::{
     ffi::{CString, c_char, c_void},
     mem, slice,
 };
+use unleash_types::client_metrics::MetricBucket;
 mod logging;
 
 #[allow(clippy::all)]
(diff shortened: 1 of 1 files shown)


pure-wasm/src/lib.rs
```diff
@@ -113,43 +119,60 @@ pub extern "C" fn dealloc_response_buffer(ptr: *mut u8, len: usize) {
     }
 }
 
-// This is only used as a thread local so it's thread safe. If this gets moved out of the
-// thread local this will ruin someone's day. So don't do that. Please bro
```
sighphyre, line 117: Making this comment go away is basically the point of this PR

pure-wasm/src/lib.rs
```diff
@@ -113,43 +119,60 @@ pub extern "C" fn dealloc_response_buffer(ptr: *mut u8, len: usize) {
...
-            if let Some(flag) = enabled {
-                resp_builder.add_enabled(flag);
-                resp_builder.add_has_enabled(true);
-            }
-            if let Some(err) = error_offset {
-                resp_builder.add_error(err);
-            }
-            resp_builder.finish()
-        };
+            let offset = Self::as_flat_buffer(&mut builder, input);
```
sighphyre, line 130: We use this top level implemented function on the trait, which dispatches to the user defined trait impl. This means the surrounding lines of code here can't be forgotten so long as you follow the pattern

pure-wasm/src/lib.rs
```diff
@@ -113,43 +119,60 @@ pub extern "C" fn dealloc_response_buffer(ptr: *mut u8, len: usize) {
...
+            builder.finished_data().to_vec()
+        })
+    }
 }
 
-pub fn build_variant_response(variant: Option<ExtendedVariantDef>, error: Option<&str>) -> Vec<u8> {
-    BUILDER.with(|cell| {
-        let mut builder = cell.borrow_mut();
-        builder.reset();
+impl FlatbufferSerializable<Result<Option<bool>, &str>> for Response<'static> {
```
sighphyre, line 138: Same code really, just handling errors more honestly

daveleek (approved):
You might want @chriswk' review as well, but this is cool! LG

PR #248, opened 2025-05-06
refactor: move some rust code into its own files and use a tryfrom fo…

sighphyre:
This is mostly just shuffling code around. This does two things

1) Gives a bunch of Rust code its own home in its own files so they don't pollute the public API
2) Moves the context materialization code to a single `try_from` so that isn't duplicated in both `check_enabled` and `check_variant`

### DIFF
--- a/pure-wasm/src/random.rs
+++ b/pure-wasm/src/random.rs
@@ -1,0 +1,20 @@
+#[unsafe(no_mangle)]
+pub fn get_random() -> i32 {
+    rand::random::<i32>()
+}
+
+//This is expected to be defined by the caller and passed to the WASM layer
+unsafe extern "C" {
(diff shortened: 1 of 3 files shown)


pure-wasm/src/random.rs
```diff
@@ -0,0 +1,20 @@
+#[unsafe(no_mangle)]
```
sighphyre, line 1: Not new, just extracted from lib.rs

pure-wasm/src/serialisation.rs
```diff
@@ -0,0 +1,176 @@
+use std::cell::RefCell;
```
sighphyre, line 1: Also not new, also extracted from lib.rs

daveleek (approved):
Nice!

PR #249, opened 2025-05-06
chore: plumbing for custom strategies

sighphyre:
Does nothing on its own. This is just plumbing to be able to pass custom strategies results like they used to be passed


java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
```diff
@@ -0,0 +1,190 @@
...
+    if (this.registeredStrategies.isEmpty() && this.fallbackStrategy == null) {
+      return;
+    }
+
+    if (toggles == null || toggles.isEmpty()) {
+      return;
+    }
+
+    try {
+      VersionedFeatures wrapper =
```
daveleek, line 56: What's the "versioned" part referring to?

daveleek (approved):
Cool!

java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
sighphyre: That's a good question and I actually don't know. I didn't write that, just uncommenting it

sighphyre: Sorry, undeleting it

PR #250, opened 2025-05-06
chore: restore custom strategies

sighphyre:
Restores the custom strategies code from main and fixes the tests


java-engine/src/main/java/io/getunleash/engine/CustomStrategiesEvaluator.java
```diff
@@ -17,7 +17,7 @@
 
 class CustomStrategiesEvaluator {
   private static final Logger log = LoggerFactory.getLogger(CustomStrategiesEvaluator.class);
-  static final String EMPTY_STRATEGY_RESULTS = "{}";
+  static final Map<String, Boolean> EMPTY_STRATEGY_RESULTS = Map.of();
```
sighphyre, line 20: The public API here now returns a Map<String, Boolean> instead of a JSON string, which isn't helpful for flat buffers

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -196,6 +204,16 @@ public UnleashEngine() {
 
     ExportFunction newEngine = instance.export("new_engine");
 
+    if (customStrategies != null && !customStrategies.isEmpty()) {
```
sighphyre, line 207: This block is stolen from main, basically the same thing

PR #251, opened 2025-05-06
feat: wasm side to custom strategies

sighphyre:
Adds the Rust side to custom strategies, a few bug fixes and restores the original tests


java-engine/src/main/java/io/getunleash/engine/Context.java
```diff
@@ -39,4 +39,8 @@ public String getRemoteAddress() {
   public Map<String, String> getProperties() {
     return properties;
   }
+
+  public void setProperties(Map<String, String> properties) {
```
sighphyre, line 43: Just so the tests can use this

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -95,8 +95,8 @@ private static int[] buildCustomStrategyResults(
         continue;
       }
       int keyOffset = builder.createString(entry.getKey());
-      int valueOffset = builder.createString(entry.getValue().toString());
-      int propOffset = PropertyEntry.createPropertyEntry(builder, keyOffset, valueOffset);
+      int propOffset =
+          PropertyEntry.createPropertyEntry(builder, keyOffset, entry.getValue() ? 1 : 0);
```
sighphyre, line 99: Embarrassing bug fix that totally didn't take me 20 minutes to find. If you send a String as a boolean then the first bit tends to be set and the Rust side tends to see that as a `true`. So instead you should just pass booleans as booleans, which in practice, is a single bit

PR #253, opened 2025-05-06
feat: core version and built in strategies

sighphyre:
Implements core version and built in strategies. This is done by making the WASM setup a static initializer block so that we can call these two new methods statically, rather than requiring an instance


java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
```diff
@@ -286,14 +286,30 @@ void customStrategiesRequired_whenNotConfigured_returnsFalse(
...
+    assertTrue(strategies.contains("gradualRolloutUserId"));
+    assertTrue(strategies.contains("gradualRolloutRandom"));
+    assertTrue(strategies.contains("applicationHostname"));
+    assertTrue(strategies.contains("gradualRolloutSessionId"));
+    assertTrue(strategies.contains("remoteAddress"));
+    assertTrue(strategies.contains("flexibleRollout"));
+  }
+
+  @Test
+  void testCoreVersionIsRetrieved() {
```
sighphyre, line 306: This is restored from main. It's not new

java-engine/src/test/java/io/getunleash/engine/UnleashEngineTest.java
```diff
@@ -286,14 +286,30 @@ void customStrategiesRequired_whenNotConfigured_returnsFalse(
...
-  // @Test
-  // void testCoreVersionIsRetrieved() {
-  // String coreVersion = UnleashEngine.getCoreVersion();
-  // assertNotNull(coreVersion);
-  // // check that it contains two dots, close enough for a quick and dirty but
-  // // stable semver check
-  // assertTrue(coreVersion.split("\\.").length >= 3);
-  // }
+  @Test
+  void testBuiltInStrategiesAreRetrieved() {
```
sighphyre, line 290: This is new, it's a bit dumb but I think it's okay

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -34,20 +36,66 @@
...
-  private ExportFunction getMetrics;
-  private ExportFunction deallocResponseBuffer;
-  private ExportFunction getLogBufferPtr;
-  private ExportFunction listKnownToggles;
   private Memory memory;
-
   private final CustomStrategiesEvaluator customStrategiesEvaluator;
 
+  static {
+    ImportValues imports =
```
sighphyre, line 56: Ripped out of the constructor. This gives us a path to accessing WASM functions without having an UnleashEngine object, which is nice for static data like the version

PR #260, opened 2025-05-27
chore: test python build

sighphyre:
Poetry got updated to 2.x. That version doesn't package native library correctly with out build scripts, which leads to a package that crashes on startup. This pins the Poetry version in the build to a working version


gastonfournier (approved):
LG!

PR #262, opened 2025-05-28
fix: lock the ffi layer so that it can be safely accessed

sighphyre:
Wraps the FFI layer's engine state in an atomic reference counted pointer and a read/write lock so that it can be safely accessed a multiple threads at the same time


yggdrasilffi/src/lib.rs
```diff
@@ -115,6 +112,25 @@ fn result_to_json_ptr<T: Serialize>(result: Result<Option<T>, FFIError>) -> *mut
...
 }
 
+unsafe fn get_engine(engine_ptr: *mut c_void) -> Result<ManagedEngine, FFIError> {
+    if engine_ptr.is_null() {
+        return Err(FFIError::NullError);
+    }
+    let arc_instance = Arc::from_raw(engine_ptr as *const RawPointerDataType);
+
+    let cloned_arc = arc_instance.clone();
+    forget(arc_instance);
```
sighphyre, line 122: ...but now we have a problem because we've reconstituted the Arc from the raw pointer and when this function exits, it'll be dropped (Crab-Speak for garbage collected).

So we tell the Rust compile to forget about it (please don't clean it up).

This is super important because without this it's possible to hit 0 references to this and have it dropped which means the whole program catches fire

yggdrasilffi/src/lib.rs
```diff
@@ -115,6 +112,25 @@ fn result_to_json_ptr<T: Serialize>(result: Result<Option<T>, FFIError>) -> *mut
...
+
+    Ok(cloned_arc)
+}
+
+fn recover_lock<T>(lock: &Mutex<T>) -> MutexGuard<T> {
+    match lock.lock() {
+        Ok(guard) => guard,
+        Err(poisoned) => poisoned.into_inner(),
```
sighphyre, line 130: A Mutex becomes poisoned when a thread panics while holding a lock to it. From Yggdrasil's perspective, either this means

- Everything is on fire and nothing will work, in which case re-accessing the poisoned Mutex won't make it worse
- Something panicked while accessing something that is safe to recover from and we can ignore the error

So that's what we do in both cases - acknowledge the poisoned state and then ignore it, by recovering with `into_inner()`

yggdrasilffi/src/lib.rs
```diff
@@ -123,8 +139,9 @@ fn result_to_json_ptr<T: Serialize>(result: Result<Option<T>, FFIError>) -> *mut
 /// `free_engine` and passing in the pointer returned by this method. Failure to do so will result in a leak.
 #[no_mangle]
 pub extern "C" fn new_engine() -> *mut c_void {
-    let engine = EngineState::default();
-    Box::into_raw(Box::new(engine)) as *mut c_void
+    let engine = Mutex::new(EngineState::default());
```
sighphyre, line 142: This function is a the salient change. Rather than just a heap value, it's now a heap value wrapped in a thread safe read lock, which is wrapped in a reference counting pointer

PR #265, opened 2025-06-05
chore: internal refactor to free yggdrasil from random/sys clock, move to ahash and optimize flow of is_enabled/get_variant

sighphyre:
This implements some internal optimiztions for the Yggdrasil and FFI layers so that this can be compiled to WASM effectively

- Makes all IO optional, since WASM can't do IO. System clock is now behind a "wall-clock" compiler flag, hostname can now be passed in the context
- Moved to AHash, which yielded some modest perf improvements when moving to WASM
- is_enabled/get_variant now take the EnrichedContext struct directly rather than assembling it internally. This allows us to pass that all the way down from the FFI layer/WASM layer without packing and unpacking structs along a computationally bound path
- For metrics, the current timestamp is now passed from external sources. For the FFI layer, that's simply calculated in the FFI crate


chriswk (approved):
LGTM

PR #266, opened 2025-06-05
feat: implement pure wasm layer

sighphyre:
Adds a pure WASM (not WASI) layer to Yggdrasil. This uses flatbuffers for an efficient serialization protocol, and unlike the FFI layer, does is_enabled/get_variant in one shot - i.e. metrics and impression data are included as part of the flow.

The contents of enabled-message_generated.rs are autogenerated by flatc and can be ignored for review purposes

This probably needs some tests, coming in a later PR


chriswk (approved):
LGTM. Waiting with excitement for the tests :P

PR #267, opened 2025-06-05
feat: pure wasm java engine

sighphyre:
Moves the Java engine to use a pure WASM implementation


java-engine/build.gradle.kts
```diff
@@ -2,10 +2,25 @@ plugins {
...
     id("tech.yanand.maven-central-publish").version("1.3.0")
+    id("me.champeau.jmh").version("0.7.2")
+    id("at.released.wasm2class.plugin").version("0.3")
+}
+
+wasm2class {
+    modules {
+        create("Yggdrasil") {
+            wasm = file("../target/wasm32-unknown-unknown/release/pure_wasm.wasm")
+            targetPackage = "org.example.wasm"
```
chriswk, line 17: Seems like the wrong targetPackage? I'd think io.getunleash.wasm would be better?

PR #270, opened 2025-06-06
docs: use and build the wasm core

sighphyre:


pure-wasm/README.md
```diff
@@ -1,5 +1,36 @@
...
+``` bash
+flatc --rust -o pure-wasm/src flat-buffer-defs/enabled-message.fbs
+```
+
+### Building
+
+This project uses low-level bit-hacking optimizations that rely on a 32-bit architecture, so the only supported target is wasm32-unknown-unknown. Build configuration is already defined in the root project's config.toml, so you can compile the WASM module with:
+
+``` bash
+cargo build --release --target wasm32-unknown-unknown
```
gastonfournier, line 25: Not sure if this is or not an important detail. Rust compiler still gives you this command if it fails due to the missing target
```suggestion
rustup target add wasm32-unknown-unknown # only required once
cargo build --release --target wasm32-unknown-unknown
```

sighphyre: Sure we can add this. But lemme do it earlier

PR #272, opened 2025-06-06
docs: readme docs for java engine

sighphyre:
Adds some docs for the Java Yggdrasil engine

Does a sneaky thing to the build too so that you don't have to manually compile the WASM engine


chriswk (approved):
Yeah. Once I realised this wasn't our public SDK docs... :)

PR #274, opened 2025-06-09
fix(java): don't warn when using a custom strategy on built in strategies

sighphyre:
Built in strategies are now hydrated into the engine so that it doesn't complain when it can't detect "flexibleRollout" as a built in strategy

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
+++ b/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
@@ -49,7 +49,7 @@ public class UnleashEngine {
       IStrategy fallbackStrategy,
       NativeInterface nativeInterface) {
     if (customStrategies != null && !customStrategies.isEmpty()) {
-      List<String> builtInStrategies = new ArrayList<>();
+      List<String> builtInStrategies = getBuiltInStrategies();
       this.customStrategiesEvaluator =
           new CustomStrategiesEvaluator(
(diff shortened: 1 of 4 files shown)


unleash-yggdrasil/src/lib.rs
```diff
@@ -566,7 +566,7 @@ impl EngineState {
             .map(|seed| {
                 normalized_hash(group_id, &seed, total_weight, VARIANT_NORMALIZATION_SEED).unwrap()
             })
-            .unwrap_or_else(|| rand::thread_rng().gen_range(0..total_weight));
+            .unwrap_or_else(|| rand::rng().random_range(0..total_weight));
```
sighphyre, line 569: Just some boyscouting - old method got deprecated recently and this is a drop in replacement

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -49,7 +49,7 @@ public UnleashEngine(
       IStrategy fallbackStrategy,
       NativeInterface nativeInterface) {
     if (customStrategies != null && !customStrategies.isEmpty()) {
-      List<String> builtInStrategies = new ArrayList<>();
+      List<String> builtInStrategies = getBuiltInStrategies();
```
gastonfournier, line 52: So this evaluates all custom and built in strategies (as it's Java bytecode now)

gastonfournier (approved):
LG!

PR #276, opened 2025-06-12
fix: patch take state encoding

sighphyre:
Windows likes to use things that are not UTF-8. That's bad. Flatbuffers will enforce UTF-8 across the WASM boundary. That's good. We're not using Flatbuffers for takeState. That's bad.

So we can just force the JVM to read this with an encoding we choose


chriswk (approved):
LGTM.

PR #286, opened 2025-07-03
Java: fix get variant counts

mturlo:
Hi! After updating to the WASM-based Unleash Java Client 11.0.0 I've noticed a strange behaviour - all the exposure metrics collected via `getVariant` (and, by extension, `evaluateAllToggles`) were counted as a "yes" - even when the strategies resolved the toggle to not exposed. I think I got the bug - I've tweaked the `testMetrics()` case slightly to expose this case.

## About the changes
The changes contain:
- modified test scenario to replicate my case
- one-liner fix in the `pure_wasm` module
- some code cleanup in `UnleashEngineTest.java`


sighphyre (approved):
Thank you so much, gosh that's a silly bug

sighphyre:
@mturlo Thanks again for the patch! This is released in Java SDK 11.0.2

PR #295, opened 2025-08-13
feat: expose get_state method in core binaries

kwasniew:
## About the changes

Closes #

### DIFF
--- a/pure-wasm/src/lib.rs
+++ b/pure-wasm/src/lib.rs
@@ -192,6 +192,22 @@ pub extern "C" fn take_state(engine_ptr: u32, json_ptr: u32, json_len: u32) {
 }
 
 #[unsafe(no_mangle)]
+pub extern "C" fn get_state(engine_ptr: u32) -> u32 {
+    let lock = unsafe { get_engine(engine_ptr as *const u32).unwrap() };
+    let engine = recover_lock(&lock);
+
+    let state = engine.get_state();
+    if let Ok(json_str) = serde_json::to_string(&state) {
+        let json_bytes = json_str.as_bytes();
+        let json_ptr = json_bytes.as_ptr() as u32;
+        std::mem::forget(json_str); // Prevent automatic deallocation
+        return json_ptr;
+    }
+
+    0 // Return null pointer only if serialization failed (shouldn't happen)
+}
+
+#[unsafe(no_mangle)]
 pub extern "C" fn check_enabled(engine_ptr: i32, message_ptr: i32, message_len: i32) -> u64 {
     let enabled: Result<ResponseMessage<bool>, WasmError> = (|| {
         let bytes =
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -658,6 +658,14 @@ impl EngineState {
         }
     }
 
+    pub fn get_state(&self) -> ClientFeatures {
+        if self.compiled_state.is_some() {
+            self.previous_state.clone()
+        } else {
+            ClientFeatures::default()
+        }
+    }
+
     pub fn apply_client_features(&mut self, toggles: ClientFeatures) -> Option<Vec<EvalWarning>> {
         let (compiled_state, warnings) = compile_state(&toggles);
         self.previous_state = toggles;
--- a/yggdrasilffi/src/lib.rs
+++ b/yggdrasilffi/src/lib.rs
@@ -10,6 +10,7 @@ use std::{
 use chrono::Utc;
 use libc::c_void;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use unleash_types::client_features::ClientFeatures;
 use unleash_types::client_metrics::MetricBucket;
 use unleash_yggdrasil::{
     state::EnrichedContext, Context, EngineState, EvalWarning, ExtendedVariantDef,
@@ -189,6 +190,23 @@ pub unsafe extern "C" fn take_state(
         } else {
             Ok(Some(()))
         }
+    })();
+
+    result_to_json_ptr(result)
+}
+
+/// Gets the current state of the engine as a JSON encoded `ClientFeatures` structure.
+///
+/// # Safety
+///
+/// This function dereferences the engine_ptr argument, and so this function should not
+/// be called with a null pointer.
+#[no_mangle]
+pub unsafe extern "C" fn get_state(engine_ptr: *mut c_void) -> *const c_char {
+    let result: Result<Option<ClientFeatures>, FFIError> = (|| {
+        let guard = get_engine(engine_ptr)?;
+        let engine = recover_lock(&guard);
+        Ok(Some(engine.get_state()))
     })();
 
     result_to_json_ptr(result)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,186 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+Yggdrasil is a multi-language SDK core for Unleash feature flags, written in Rust and providing bindings for multiple programming languages. The core Rust library (`unleash-yggdrasil`) implements the domain logic for feature flag evaluation, which is then exposed to various language ecosystems through FFI or WASM.
+
+## Architecture
+
+The project follows a hub-and-spoke architecture:
+
+- **Core Rust Library** (`unleash-yggdrasil/`): Contains the main feature evaluation logic, strategy parsing, state management, and compiled toggle functionality
+- **FFI Layer** (`yggdrasilffi/`): Provides C-compatible FFI bindings for native language integrations
+- **WASM Layer** (`yggdrasilwasm/`, `pure-wasm/`): Compiles the core to WebAssembly for JavaScript/browser usage
+- **Language Bindings**: Each language directory contains specific bindings and wrappers:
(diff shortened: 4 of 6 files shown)


pure-wasm/src/lib.rs
```diff
@@ -191,6 +191,22 @@ pub extern "C" fn take_state(engine_ptr: u32, json_ptr: u32, json_len: u32) {
...
+#[unsafe(no_mangle)]
+pub extern "C" fn get_state(engine_ptr: u32) -> u32 {
+    let lock = unsafe { get_engine(engine_ptr as *const u32).unwrap() };
+    let engine = recover_lock(&lock);
+
+    let state = engine.get_state();
+    if let Ok(json_str) = serde_json::to_string(&state) {
+        let json_bytes = json_str.as_bytes();
+        let json_ptr = json_bytes.as_ptr() as u32;
+        std::mem::forget(json_str); // Prevent automatic deallocation
```
kwasniew, line 203: This function intentionally leaks memory (the JSON string) to make it available to the calling code.

pure-wasm/src/lib.rs
```diff
@@ -191,6 +191,22 @@ pub extern "C" fn take_state(engine_ptr: u32, json_ptr: u32, json_len: u32) {
...
 
+#[unsafe(no_mangle)]
+pub extern "C" fn get_state(engine_ptr: u32) -> u32 {
+    let lock = unsafe { get_engine(engine_ptr as *const u32).unwrap() };
+    let engine = recover_lock(&lock);
+
+    let state = engine.get_state();
+    if let Ok(json_str) = serde_json::to_string(&state) {
+        let json_bytes = json_str.as_bytes();
+        let json_ptr = json_bytes.as_ptr() as u32;
```
kwasniew, line 202: Returns the memory address of the JSON string (as a u32)

pure-wasm/src/lib.rs
```diff
@@ -191,6 +191,22 @@ pub extern "C" fn take_state(engine_ptr: u32, json_ptr: u32, json_len: u32) {
     }
 }
 
+#[unsafe(no_mangle)]
+pub extern "C" fn get_state(engine_ptr: u32) -> u32 {
```
kwasniew, line 195: I noticed we don't have any tests in pure-wasm and all functions are tested in language bindings. I will do the same for this one

yggdrasilffi/src/lib.rs
```diff
@@ -194,6 +195,23 @@ pub unsafe extern "C" fn take_state(
...
+/// # Safety
+///
+/// This function dereferences the engine_ptr argument, and so this function should not
+/// be called with a null pointer.
+#[no_mangle]
+pub unsafe extern "C" fn get_state(engine_ptr: *mut c_void) -> *const c_char {
+    let result: Result<Option<ClientFeatures>, FFIError> = (|| {
+        let guard = get_engine(engine_ptr)?;
+        let engine = recover_lock(&guard);
+        Ok(Some(engine.get_state()))
```
sighphyre, line 209: Haven't tried it but I want to say if we write this to a JSON string on this line we can return a string here and then the host binding languages don't have to write that back to a string.

More a matter of taste than anything else. This isn't a hot path so the perf doesn't matter here and this approach will absolutely work. It would be nice if the WASM and FFI layer did the same thing though

unleash-yggdrasil/src/lib.rs
```diff
@@ -2106,4 +2114,45 @@ mod test {
...
+                    name: "default".into(),
+                    constraints: None,
+                    parameters: None,
+                    segments: None,
+                    sort_order: None,
+                    variants: None,
+                }]),
+                ..Default::default()
+            }],
+            version: 3,
```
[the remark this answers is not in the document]
sighphyre: Is this actually a new version? I don't remember a v3 schema floating around, what changed?

ruby-engine/spec/yggdrasil_engine_spec.rb
```diff
@@ -25,6 +25,22 @@ def test_suite_variant(base_variant)
     end
   end
 
+  describe '#getting state' do
```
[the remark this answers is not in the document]
sighphyre: No this is probably a missing build action, thanks for raising this

ruby-engine/build.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-cargo build --release
```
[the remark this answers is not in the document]
sighphyre: Thank you! Out of curiosity, what what the error?

ruby-engine/lib/yggdrasil_engine.rb
```diff
@@ -85,6 +86,13 @@ def take_state(toggles)
     YggdrasilEngine.free_response(response_ptr)
   end
 
+  def get_state
```
sighphyre, line 89: This looks subtly off, the Rust layer is returning a wrapped response that also includes a status code. That looks a little like this

``` json
{"status_code":"Ok","value":{"version":1,"features":[{"name":"testFeature","enabled":true,"strategies":[{"name":"default","variants":[]}]}],"query":null},"error_message":null}
```

You'd need to parse the JSON response and return the `value` here and/or make a decision around returning an error. Might be worth taking a peek at `get_variant` or `is_enabled` to see how they handle that dance

ruby-engine/spec/yggdrasil_engine_spec.rb
```diff
@@ -25,6 +25,22 @@ def test_suite_variant(base_variant)
...
+    it 'should return JSON and support roundtrip' do
+       empty_state = yggdrasil_engine.get_state
+       expect(empty_state).to include('"features":[]')
+
+       test_state = '{"version":1,"features":[{"name":"testFeature","enabled":true,"strategies":[{"name":"default"}]}]}'
+       yggdrasil_engine.take_state(test_state)
+       retrieved_state = yggdrasil_engine.get_state
+
+       expect(retrieved_state).to include('testFeature')
```
sighphyre, line 37: Ah that makes sense. These tests are burning us here because the response contains the wrapping so it's not correct but does these individual properties.

Damnit, that's quite an easy and subtle trap to fall into. Would love your thoughts on how we can make this a harder thing to do by mistake

python-engine/yggdrasil_engine/engine.py
```diff
@@ -182,6 +185,11 @@ def take_state(self, state_json: str) -> Optional[List[Warning]]:
                 return warnings
             return None
 
+    def get_state(self) -> str:
+        response_ptr = self.lib.get_state(self.state)
+        with self.materialize_pointer(response_ptr, dict) as result:
+            return json.dumps(result.value)
```
sighphyre, line 191: Same as Ruby I think we'll want to unwrap the wrapping, so returning `result['value']` is probably what we want

python-engine/build.sh
```diff
@@ -1,4 +1,4 @@
-cargo build --release
+cargo build --release -p yggdrasilffi
```
sighphyre, line 1: :heart:

pure-wasm/src/lib.rs
```diff
@@ -191,6 +191,22 @@ pub extern "C" fn take_state(engine_ptr: u32, json_ptr: u32, json_len: u32) {
...
+    let lock = unsafe { get_engine(engine_ptr as *const u32).unwrap() };
+    let engine = recover_lock(&lock);
+
+    let state = engine.get_state();
+    if let Ok(json_str) = serde_json::to_string(&state) {
+        let json_bytes = json_str.as_bytes();
```
sighphyre, line 201: I'd rather not go through the dance of manually assembling CStrings here. It's weirdly hard to get right. Instead we can just use Rust's built in FFI handling stuff, which should work equally well for WASM. We can replace the last 3 lines of this function with

```
        let c_string = CString::new(json_bytes).unwrap();
        return c_string.into_raw() as u32
```

pure-wasm/src/lib.rs
sighphyre: Yeah, this is kinda intentional. Most of this is wrapping and the only real way to test this is to setup a Rust FFI reader for this and poke it in the same way the host would poke it. Which feels daft because we can just test this with host tests. I'm open to being told I'm crazy here and we can always set something up if we need

java-engine/src/main/java/io/getunleash/engine/WasmInterface.java
```diff
@@ -136,6 +140,14 @@ public void takeState(int enginePtr, byte[] messageBytes) {
     }
   }
 
+  @Override
+  public String getState(int enginePtr) {
+    synchronized (engineLock) {
+      int ptr = (int) getState.apply(enginePtr)[0];
+      return instance.memory().readCString(ptr);
```
sighphyre, line 147: I'm pretty sure this needs a free. I don't think just reading the string is enough to free it (if you were looking at readLog, that's intentionally not freed this side because it's freed on the Rust side).

I think you'd need a new function for this on the Rust side. Something like this

```
pub extern "C" fn free_cstring(ptr: *mut c_char) {
    if !ptr.is_null() {
        unsafe {
            let _ = CString::from_raw(ptr);
        }
    }
}
```

The FFI layer should have a function called free_response which should do what you want.

sighphyre (commented):
Some traps here that probably need some attention but I'm impressed! Well done dude!

ruby-engine/build.sh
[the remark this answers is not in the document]
sighphyre: Ah, little bugger isn't respecting its path and is trying to build all the projects in the subfolder

PR #296, opened 2025-08-14
chore: add tests validating the failure scenario

gastonfournier:
## About the changes
Based on the report https://github.com/Unleash/unleash-java-sdk/issues/319 we've identified a few places where the engine could fail and runtime exceptions may bubble up making the SDK unreliable.

This PR adds tests for those cases and makes the required changes to the engine

### DIFF
--- a/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
+++ b/java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
@@ -1,7 +1,18 @@
 package io.getunleash.engine;
 
-import com.dylibso.chicory.runtime.TrapException;
 import com.google.flatbuffers.FlatBufferBuilder;
+import messaging.BuiltInStrategies;
+import messaging.ContextMessage;
+import messaging.FeatureDefs;
+import messaging.MetricsResponse;
+import messaging.PropertyEntry;
+import messaging.Response;
+import messaging.ToggleEntry;
+import messaging.ToggleStats;
+import messaging.Variant;
+import messaging.VariantEntry;
+import messaging.VariantPayload;
+
 import java.lang.ref.Cleaner;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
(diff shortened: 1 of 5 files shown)


java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -192,110 +188,129 @@ private static byte[] buildMessage(
...
-    Response response = nativeInterface.checkEnabled(enginePointer, contextBytes);
+      Response response = nativeInterface.checkEnabled(enginePointer, contextBytes);
 
-    if (response.error() != null) {
-      String error = response.error();
-      throw new YggdrasilInvalidInputException(error);
-    }
+      if (response.error() != null) {
+        String error = response.error();
+        throw new YggdrasilInvalidInputException(error);
```
augustyn-tink, line 234: You are throwing an exception here, that is caught in line 244. Doesn't make much sense, as it will be inlined by jvm either way. Why not return the `WasmIsEnabledResponse(false,null)` here, having logged the error message before, and leave the try-catch for real unexpected errors?

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -192,110 +188,129 @@ private static byte[] buildMessage(
...
+        if (variantPayload != null) {
+          payload = new Payload();
+          payload.setType(variant.payload().payloadType());
+          payload.setValue(variant.payload().value());
+        }
+
+        if (variant.error() != null) {
+          String error = variant.error();
+          throw new YggdrasilInvalidInputException(error);
+        }
```
augustyn-tink, line 269: This check could be moved to just after the line 254, so JVM won't have to unnecessarily assign few bytes of memory in case of error, and to align with "fail fast" approach.

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
sighphyre: I don't actually think this will be caught in that catch block. YggdrasilInvalidInputException inherits from the base Exception class not RuntimeException. I think this is okay in this case, YggdrasilInvalidInputException is a "oh gosh something went wrong but we know what to do about it" whereas the RuntimeException is catching stuff from Chicory that's a much, much worse case here.

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -192,110 +188,129 @@ private static byte[] buildMessage(
...
+  public void takeState(String clientFeatures) throws YggdrasilInvalidInputException {
     try {
       customStrategiesEvaluator.loadStrategiesFor(clientFeatures);
       byte[] messageBytes = clientFeatures.getBytes(StandardCharsets.UTF_8);
       nativeInterface.takeState(this.enginePointer, messageBytes);
-    } catch (TrapException e) {
-      throw e;
+    } catch (RuntimeException e) {
+      throw new YggdrasilInvalidInputException(
+          "Failed to take state, ensure the input is a valid Yggdrasil state file", e);
```
sighphyre, line 198: Changing this here because this isn't how format exceptions get bubbled. This should only occur if something goes wrong deep in Chicory
```suggestion
          "Failed to take state:", e);
```

sighphyre (approved):
Okay going to merge this I think. I've made some small changes to the branch but nothing massive I think. Welcome to come poke me and we can reintroduce the WasmResponse changes in another branch if you're desperate for them but I have strong feelings here

PR #297, opened 2025-08-14
chore: upgrade chicory library and flatbuffers

gastonfournier:
## About the changes
Bumps two dependencies in which this engine relies on.

### DIFF
--- a/java-engine/gradle.properties
+++ b/java-engine/gradle.properties
@@ -1,3 +1,3 @@
 group=io.getunleash
 yggdrasilCoreVersion=0.17.3
-version=0.3.2
+version=0.4.0


sighphyre:
Oh no, flatc is a problem here because the generated files don't work with Rust. I think a test should have caught that though so that's on me.

Upgrading Chicory is awesome, I'll see if I can pull these two apart and add a test that catches the flatc stuff

java-engine/src/main/java/io/getunleash/engine/UnleashEngine.java
```diff
@@ -194,7 +194,8 @@ public void takeState(String clientFeatures) throws YggdrasilInvalidInputExcepti
       byte[] messageBytes = clientFeatures.getBytes(StandardCharsets.UTF_8);
       nativeInterface.takeState(this.enginePointer, messageBytes);
     } catch (RuntimeException e) {
-      throw new YggdrasilInvalidInputException("Failed to take state:", e);
+      throw new YggdrasilInvalidInputException(
```
sighphyre, line 197: I don't think this is correct here. The passed data being an invalid state file won't trigger this exception, this only occurs if something goes wrong with the internals of Chicory

java-engine/src/main/java/io/getunleash/engine/WasmIsEnabledResponse.java
```diff
@@ -0,0 +1,7 @@
+package io.getunleash.engine;
+
+public class WasmIsEnabledResponse extends WasmResponse<Boolean> {
```
sighphyre, line 3: I don't understand what this adds. From what I can tell it's just abstraction for abstraction's sake. Can we roll this back unless it adds something meaningful? The original generic response encodes what it needs in the type signature without hiding information

java-engine/README.md
```diff
@@ -118,9 +118,14 @@ The WASM library is automatically built on running tests but you'll need to make
...
 
+**Note:** assume flatc is available in your PATH, if not, you can specify the full path to flatc in the command below.
+
 ``` bash
+cd .. # go to yggdrasil root directory
+rustup target add wasm32-unknown-unknown
+cargo build --release --target wasm32-unknown-unknown
 flatc --java -o java-engine/src/main/java flat-buffer-defs/enabled-message.fbs
-
+cd pure-wasm && cargo build --release --target wasm32-unknown-unknown
```
sighphyre, line 128: This shouldn't be needed, the gradle build does this step behind the scenes

java-engine/README.md
```diff
@@ -118,9 +118,14 @@ The WASM library is automatically built on running tests but you'll need to make
 
 The Java engine uses FlatBuffers for communication with the WASM core. If you make changes to the data interchange format, regenerate the bindings like this:
 
+**Note:** assume flatc is available in your PATH, if not, you can specify the full path to flatc in the command below.
+
 ``` bash
+cd .. # go to yggdrasil root directory
+rustup target add wasm32-unknown-unknown
```
sighphyre, line 125: I'd rather take these changes to the README out. Further up in the README, there's instructions to go complete the setup for the WASM engine, which already includes these steps. I'd rather have a single source of truth for this information so that this doesn't drift. If that needs patches can we make the changes there rather and/or try to make it clearer that you should go and do those steps first?

chriswk:
Moot / superseded by moving to https://github.com/Unleash/yggdrasil-bindings.

PR #300, opened 2025-08-19
chore: bump release versions for python, ruby, java and dotnet

sighphyre:


python-engine/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "yggdrasil-engine"
-version = "0.1.6"
+version = "1.0.0"
```
kwasniew, line 3: why not 1.1.0 as ruby?

sighphyre: I see what you mean now that I look this over again though. It's suspect that 2 other engines are being bumped to 1.1.0 here. Promise it's a weird coincidence :laughing:

kwasniew: ah right. I got confused by weirdly similar numbers :D

PR #306, opened 2025-09-05
fix: add a musl build that doesn't rely on libgcc

sighphyre:
## About the changes

Closes #

### DIFF
--- a/.github/workflows/build-binaries.yaml
+++ b/.github/workflows/build-binaries.yaml
@@ -31,12 +31,10 @@ jobs:
             target: x86_64-unknown-linux-musl
             output: libyggdrasilffi.so
             name: libyggdrasilffi_x86_64-musl.so
-            cross: true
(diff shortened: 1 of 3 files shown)


.github/workflows/build-binaries.yaml
```diff
@@ -31,11 +31,13 @@ jobs:
             target: x86_64-unknown-linux-musl
             output: libyggdrasilffi.so
             name: libyggdrasilffi_x86_64-musl.so
+            compiler: x86_64-linux-musl-gcc
             cross: true
           - os: ubuntu-latest
             target: aarch64-unknown-linux-musl
             output: libyggdrasilffi.so
             name: libyggdrasilffi_arm64-musl.so
+            compiler: aarch64-linux-musl-gcc
```
chriswk, line 40: I thought it was to not rely on libgcc? your compiler choice sounds like it does depend on it :)

PR #307, opened 2025-09-09
chore: remove language bindings

sighphyre:
Removes the language bindings for Yggdrasil. These are now broken out to https://github.com/Unleash/yggdrasil-bindings so we can have independent release pipelines for the FFI/WASM layers vs the published Rust core


chriswk (approved):
Hard to review deleted lines :) Have a :+1:

PR #308, opened 2025-09-09
chore: add release plz workflow

sighphyre:
Adds a release plz workflow so we can free ourselves from cargo smart-release

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -1,0 +1,102 @@
+[workspace]
+
+changelog_update = true
+
+dependencies_update = true
+
+git_tag_enable = true
+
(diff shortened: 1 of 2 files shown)


release-plz.toml
```diff
@@ -0,0 +1,102 @@
...
+
+git_release_enable = false
+
+pr_labels = ["release"]
+
+allow_dirty = false
+
+publish_allow_dirty = false
+
+semver_check = true
```
sighphyre, line 17: This file is stolen from the Edge release-plz.toml. This is the only meaningful change I've made here because semver checks do actually make sense for Ygg

PR #339, opened 2026-03-12
feat: support semver greater than or equal and semver less than or equal

sighphyre:


sighphyre:
Merged into #340 and closed via that, this is now just duplicate work

PR #344, opened 2026-06-04
fix: avoid panic when scalar-operator constraint has null value

AbhinavGupta-de:
## Bug

`upgrade_constraint` in `strategy_upgrade.rs` unwraps `constraint.value` unconditionally in the scalar-operator branch (`NUM_*`, `DATE_*`, `SEMVER_*`). When `value` is `null` the unwrap panics. Because the panic happens inside the FFI boundary used by the language bindings (PyO3, .NET, JNI), it cannot unwind — `nounwind` aborts the entire host process, killing the worker that triggered the refresh.

## Example

Any client receiving this from `/client/features` panics on its next refresh:

```json
{
  "contextName": "createdAt",
  "operator": "DATE_AFTER",
  "value": null,
  "values": ["2024-01-01T00:00:00Z"]
}
```

One realistic way to produce this: edit the constraint in the Unleash admin UI and save a scalar into the `values[]` array instead of the `value` scalar field. The server happily stores it; the engine then aborts on the consumer side.

## Fix

Replace the unconditional `unwrap()` with a `let-else` that returns `"false"` (never-match) when `value` is `None`. This matches the existing convention in the same branch — the SemVer-with-`v`-prefix special case already returns `"false"` for broken inputs. A constraint with a missing value silently never matches instead of aborting the process.

The same guard now covers both unwrap sites in this branch (the SemVer `starts_with('v')` check and the final clone).

## Tests

Added two tests in `mod tests`:

- `scalar_operator_with_null_value_is_never_match` — parametrized over all 12 scalar operators (`NumEq`/`Lt`/`Lte`/`Gt`/`Gte`, `DateAfter`/`Before`, `SemverEq`/`Lt`/`Lte`/`Gt`/`Gte`). Each builds a constraint with `value: None` and asserts `upgrade_constraint` returns `"false"`.
- `scalar_operator_with_value_in_values_field_is_never_match` — the realistic case where the scalar landed in `values[]` instead of `value`.

Without the fix every parametrized case panics; with the fix all 13 pass. Existing tests are unaffected.

### DIFF
--- a/unleash-yggdrasil/src/strategy_upgrade.rs
+++ b/unleash-yggdrasil/src/strategy_upgrade.rs
@@ -333,21 +333,25 @@ fn upgrade_constraint(constraint: &Constraint) -> String {
             escape_quotes(constraint.value.as_ref().unwrap_or(&"".to_string()))
         )
     } else {
-        if constraint.operator == Operator::SemverEq
-            || constraint.operator == Operator::SemverLt
-            || constraint.operator == Operator::SemverGt
-            || constraint.operator == Operator::SemverLte
-            || constraint.operator == Operator::SemverGte
+        let Some(value) = constraint.value.as_ref() else {
+            return "false".into();
+        };
+        if matches!(
+            constraint.operator,
+            Operator::SemverEq
+                | Operator::SemverLt
+                | Operator::SemverGt
+                | Operator::SemverLte
+                | Operator::SemverGte
+        ) && value.starts_with('v')
         {
             // A silly special case where we want to ingest
             // broken semver operators so we can reject them.
             // Handling this in the grammar feels awful so we're
             // just not going to
-            if constraint.value.as_ref().unwrap().starts_with('v') {
-                return "false".into();
-            }
+            return "false".into();
         }
-        constraint.value.clone().unwrap()
+        value.clone()
     };
 
     format!("{inversion}{context_name} {op} {value}")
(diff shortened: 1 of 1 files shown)


sighphyre:
Hey @AbhinavGupta-de, I don't think I'm completely following your reproduction case here, the Admin UI specifically blocks this case because it would result in an invalid constraint. I want to say that that's enforced on the backend too. Have you found a way around this somehow? Could you provide more detailed steps to reproduce this?

maybeabhinav:
hey @sighphyre,  we ran into this on `yggdrasil-engine 1.0.0` against the hosted SaaS, where `/api/client/features` did serve us a `DATE_AFTER` constraint with `value: null` and the date sitting in `values[]`. Bumped to `1.3.0` and the same payload is benign now (toggle just gets dropped), so it looks like the input is already guarded upstream — probably in unleash-types or take_state.

The `.unwrap()s` themselves are still in `strategy_upgrade.rs` though, so figured it's still worth hardening at that layer too — if the upstream validation ever has a hole, or someone uses the crate directly without the Python/Java wrapper's guards, the abort comes back. Happy to close it if you'd rather keep that path strict, just thought a tiny let-else matching the SemVer-v precedent right above was cheap insurance.

sighphyre:
I feel like I've stumbled upon a murder mystery here! I can't reproduce this in Unleash itself, in theory Unleash should have never allowed that data to exist at all. I'm also reasonably sure that this doesn't have any special handling in the `take_state` calls or anywhere else. So now I don't know how the bug came to be and I don't know why it isn't present anymore :sweat_smile:

Sure this wasn't a hand crafted bootstrap or backup file in the SDK? If you do figure it out, please let us know!

That being said, if it happened once, it'll happen again so probably removing the unwrap here is a sane thing to do. I don't really want to follow the same path as the way semvers are handled - that's done like that because of silly historical reasons and I don't want to make the problem worse.

If it's okay with you, I'm going to close this, tackle it in a different way and add you as a co-author in another body of work

maybeabhinav:
Yes sure we can close this and go ahead with your implementation, I will provide you with more info on this after checking more! 

Versions:
- UnleashClient 6.4.1 (Python SDK)
- yggdrasil-engine 1.0.0 (Rust core via ctypes)

maybeabhinav:
hey @sighphyre  I just asked claude that if we could reproduce the same issue by running unleash locally and creating a small program for it. here is the reproduce, [link to github repo](https://github.com/AbhinavGupta-de/yggdrasil-null-value-panic-repro) hope this helps :)

sighphyre:
> hey @sighphyre I just asked claude that if we could reproduce the same issue by running unleash locally and creating a small program for it. here is the reproduce, [link to github repo](https://github.com/AbhinavGupta-de/yggdrasil-null-value-panic-repro) hope this helps :)

Oh you absolute gem, that's ridiculously helpful! Thank you for this! Of course - we're now trapping the panic at the FFI boundary, silly me. And it looks like our backend validation needs some love

Okay so I think that lowers the priority of this patch - still want to do it, I have something in flight but it wants a bit of refactoring first. Going to close this and tackle it a little more carefully, I'll co-author you when I have a patch I want to merge

Thank you so much for rooting around here, really amazing to get someone digging this deeply into an issue and it makes my life so much easier :heart:

PR #346, opened 2026-06-19
refactor: rework rule upgrade code so that we can avoid panics and

sighphyre:
This reworks the rule upgrade+compile path a little so that it can bubble errors instead of panicking if it encounters an error. The error cases here are pretty much all in the "oh god no why" scope of error but a panic doesn't help when this is used over an FFI boundary

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -186,7 +186,7 @@ pub fn compile(
     segment_map: &HashMap<i32, Segment>,
     warnings: &mut Vec<EvalWarning>,
 ) -> CompiledToggle {
-    let enabled_lambda = (|| {
+    let enabled_rule = (|| {
         let strategies = toggle.strategies.clone().unwrap_or_default();
         let rule_text = upgrade(&strategies, segment_map)?;
         compile_rule(rule_text.as_str())
@@ -200,7 +200,7 @@ pub fn compile(
         Box::new(|_| false)
     });
 
-    let get_variant_lambda = compile_variant_rule(toggle, segment_map).unwrap_or_else(|e| {
+    let get_variant_rule = compile_variant_rule(toggle, segment_map).unwrap_or_else(|e| {
         warnings.push(EvalWarning {
             toggle_name: toggle.name.clone(),
             message: format!("Failed to compile toggle, this will always be off {e:?}"),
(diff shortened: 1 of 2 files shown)


unleash-yggdrasil/src/lib.rs
```diff
@@ -124,13 +126,13 @@
 fn compile_variant_rule(
     toggle: &ClientFeature,
     segment_map: &HashMap<i32, Segment>,
-) -> Option<Vec<(RuleFragment, Vec<CompiledVariant>, String)>> {
+) -> Result<Option<Vec<(RuleFragment, Vec<CompiledVariant>, String)>>, SdkError> {
```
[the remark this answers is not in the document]
sighphyre: Fxied

unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -48,26 +51,28 @@
...
-    rule_text
+    Ok(rule_text)
 }
 
 pub fn build_variant_rules(
     strategies: &[Strategy],
     segment_map: &HashMap<i32, Segment>,
     toggle_name: &str,
-) -> Vec<(String, Vec<StrategyVariant>, String, String)> {
+) -> Result<Vec<(String, Vec<StrategyVariant>, String, String)>, SdkError> {
```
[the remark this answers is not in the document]
sighphyre: Fixed

unleash-yggdrasil/src/lib.rs
```diff
@@ -959,9 +975,7 @@ mod test {
     fn run_client_spec(spec_name: &str) {
         let spec = load_spec(spec_name);
         let mut engine = EngineState::default();
-        let warnings = engine.take_state(spec.state);
-
-        assert!(warnings.is_none());
```
sighphyre, line 964: Previously some client spec tests that intentionally triggered broken states would pass silently (semver operators) and short circuit to false. These now raise those warnings to the calling SDK instead

unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -7,6 +7,8 @@ use crate::state::SdkError;
 const DEFAULT_STICKINESS: &str = "user_id | session_id | random[10000]";
 const DEFAULT_RANDOM: &str = "random[10000]";
 
+pub(crate) type RawVariantRule = Vec<(String, Vec<StrategyVariant>, String, String)>;
```
sighphyre, line 10: Type's a mess, I know. This isn't worse though, just makes the Result wrapping easier to eyeball

unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -140,27 +147,24 @@ fn upgrade_strategy(
...
+                        "Unable to resolve a segment definition to a segment value".into(),
+                    ))
             })
         })
         .map(|iter| iter.collect::<Result<Vec<Vec<Constraint>>, SdkError>>())
-        .unwrap_or(Ok(vec![]));
-
-    if segments.is_err() {
-        return "false".into(); // We have a broken segment so default the entire strategy to false
-    }
```
sighphyre, line 151: This just bubbles now. Broken rules get compiled to "false" further up the chain now

gastonfournier (approved):
I quite like it!

PR #348, opened 2026-06-22
fix: variant strategy distributions with random stickiness

sighphyre:
This should fix strategy variant rollout evaluation for flexibleRollout strategies using stickiness: random.

Previously, get_variant evaluated the rollout strategy once to determine whether the toggle was enabled, then evaluated the strategy a second time to determine which strategy variant got applied. For deterministic stickiness this was harmless, but for random stickiness it produced a fresh random value on each evaluation. That effectively squared the rollout percentage for variant payloads, e.g. 20% became roughly 4%.

This changes the variant path so the matched strategy-variant rule is computed once and passed through to variant selection.

Addresses the bug raised in #345 but in a bit of a different way - intentional on my part because this approach is better for future plans around the grammar and the paths for building evaluations from them

### DIFF
--- a/unleash-yggdrasil/src/lib.rs
+++ b/unleash-yggdrasil/src/lib.rs
@@ -911,7 +911,14 @@ mod test {
     use ahash::AHashMap;
     use chrono::Utc;
     use serde::Deserialize;
-    use std::{collections::HashMap, fs};
+    use std::{
+        collections::HashMap,
+        fs,
+        sync::{
+            atomic::{AtomicUsize, Ordering},
+            Arc,
+        },
+    };
     use test_case::test_case;
     use unleash_types::client_features::{
         ClientFeaturesDelta, FeatureDependency, Override, Payload,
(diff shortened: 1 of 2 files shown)


unleash-yggdrasil/src/lib.rs
```diff
@@ -670,33 +680,40 @@ impl EngineState {
...
-                self.resolve_variant(strategy_variants.0, strategy_variants.1, context)
+        let variant = match matched_strategy {
+            Some(strategy) if strategy.has_strategy_variants() => {
+                self.resolve_variant(strategy.variants, strategy.group_id, context)
             }
-        } else {
-            self.resolve_variant(&toggle.variants, &toggle.name, context)
+            // Unleash can and will send empty lists for strategy variants when they aren't set
+            // in that case, we should treat it the same as missing variants and
+            // attempt to fall back to the top level variant
```
sighphyre, line 715: No change of logic here but I was struggling to understand why the code was handling empties and nulls until I went and rooted around in the history - so just making this nicer for future us

unleash-yggdrasil/src/lib.rs
```diff
@@ -706,10 +723,45 @@ impl EngineState {
...
     }
 
+    fn check_variant_by_toggle(
+        &self,
+        toggle: &CompiledToggle,
+        context: &EnrichedContext,
+    ) -> Option<VariantDef> {
+        let strategy_variants = self.compute_variant_strategies(toggle, context);
+        self.choose_variant(toggle, context, strategy_variants)
+    }
```
sighphyre, line 733: Pulled out into its own function so that parent enablement check can use this

unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -67,15 +67,14 @@ pub fn build_variant_rules(
...
                 custom_strat_count += 1;
             }
 
             Ok((
                 upgrade_strategy(strategy, segment_map, custom_strat_count)?,
-                strategy.variants.clone().unwrap(),
+                strategy.variants.clone().unwrap_or_default(),
                 strategy
                     .parameters
                     .as_ref()
```
sighphyre, line 80: Change in this file isn't related but it's 100% a bug. The old code here was only incrementing the custom strategy counter on strategies that contained variant strategies. So if for whatever reason you had a mix of custom strategies and strategies with variant strategies, you could encounter a case where the wrong custom strategy was computed to decide the variant strategy.

Picked up by Codex during AI code review 🤷

chriswk (approved):
LSTM - Looks sane to me

unleash-yggdrasil/src/lib.rs
```diff
@@ -41,6 +41,16 @@ pub const SUPPORTED_SPEC_VERSION: &str = "6.1.0";
...
+struct MatchedStrategyVariants<'a> {
+    variants: &'a Vec<CompiledVariant>,
+    group_id: &'a String,
+}
+
+impl MatchedStrategyVariants<'_> {
+    fn has_strategy_variants(&self) -> bool {
+        !self.variants.is_empty()
+    }
+}
```
gastonfournier, line 53: Nitpicking but if I just look at the struct and impl, I don't know what's the deal with the word `strategy`... Thinking abstractly, I believe matched variants with a group_id is more representative, regardless of whether they were matched from a strategy evaluation or something else... in other words, strategy seems like carried over contextual information that has nothing to do with the data representation, but with how the data was generated.

But maybe it serves a reason: it could be due to the way we had modeled variants in the past (global variants and environment variants). Still, once you compiled the matched variants with its group id, I think they don't need to know where they come from.

```suggestion
struct MatchedVariants<'a> {
    variants: &'a Vec<CompiledVariant>,
    group_id: &'a String,
}

impl MatchedVariants<'_> {
    fn has_variants(&self) -> bool {
        !self.variants.is_empty()
    }
}
```

gastonfournier (approved):
LGTM! Just a minor comment, which after reading the rest of the code, I think it makes sense keeping it as `MatchedStrategyVariants

PR #349, opened 2026-07-28
fix: prevent panics in bad feature config

sighphyre:
Allows strategy upgrades operating on broken date constraints to emit parse failures rather than failing

Turns on deny unwrap/expect for this file so that this can't come back in a new shape (leaves it open for tests)

### DIFF
--- a/unleash-yggdrasil/src/strategy_upgrade.rs
+++ b/unleash-yggdrasil/src/strategy_upgrade.rs
@@ -1,3 +1,5 @@
+#![deny(clippy::expect_used, clippy::unwrap_used)]
+
 use std::collections::HashMap;
 
 use unleash_types::client_features::{Constraint, Operator, Segment, Strategy, StrategyVariant};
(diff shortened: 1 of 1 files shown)


unleash-yggdrasil/src/strategy_upgrade.rs
```diff
@@ -345,11 +347,20 @@ fn upgrade_constraint(constraint: &Constraint) -> Result<String, SdkError> {
             // broken semver operators so we can reject them.
             // Handling this in the grammar feels awful so we're
             // just not going to
-            if constraint.value.as_ref().unwrap().starts_with('v') {
+            if constraint
+                .value
+                .as_ref()
+                .ok_or_else(|| {
+                    SdkError::StrategyParseError("Failed to resolve constraint value".into())
```
gastonfournier, line 354: Ok, semver should be prefixed with v... right? or do we allow things like 8.0.0?

Well, I checked and it seems like an expectation mismatch:

Image

Or we internally prepend the v (although that'd be weird)

gastonfournier: Answer was here:
```
            // A silly special case where we want to ingest
            // broken semver operators so we can reject them.
            // Handling this in the grammar feels awful so we're
            // just not going to
```

gastonfournier (approved):
LG!

